PR #36, opened 2019-12-06
Misc fixes

igormukhingmailcom:
Merging will be unblocked after changes will be manual tested


src/Resources/config/services/order_processor.xml
```diff
@@ -9,7 +9,7 @@
             <argument type="service" id="translator"/>
             <argument type="service" id="sylius.custom_factory.adjustment"/>
 
-            <tag name="sylius.order_processor"/>
+            <tag name="sylius.order_processor" priority="0"/>
```
loevgaard: I suppose `0` is default?

igormukhingmailcom: Yes, but I provided default value explicitly here just just things to be clear (that gift card processor will be running AFTER adjustments clearer that have priority=60)

loevgaard: I understand. Let's just keep it :)

PR #50, opened 2020-01-14
Add customer to gift card

loevgaard:


tests/Application/config/packages/_sylius.yaml
```diff
@@ -13,6 +13,13 @@ sylius_shop:
     product_grid:
         include_all_descendants: true
 
+
```
igormukhingmailcom: Its better to move this config and all other configs that required by gift card plugin to separate file `sylius_gift_card.yaml`

PR #56, opened 2020-03-02
[WIP] Adding a PDF generation

Roshyo:
* [x] Create a GiftCard configuration entity
* [x] Relation between Channel/Locale and Configuration
* [x] Generate PDF
* [x] Allow Admin to download it
* [x] Allow ShopUser to download it
* [x] Security about viewing PDF
* [x] Form Validation
* [x] DB Constraints
* [x] Associate Gift Card to configuration once generated
* [x] Allow a configuration to be the default one
* [x] Display information (amount/code) on the right places in PDF
* [x] Add tests/specs


igormukhingmailcom:
I have also next idea for PDF generation feature:
- Add new fields to GiftCard or GiftCardImage entity like `renderNameAtX`, `renderNameAtY`, `renderCodeAtX`, `renderCodeAtY` (or something like `nameCoords`, `codeCoords` which will contain something like `X,Y`)
- Render customer's name and gift card's code at defined coords

loevgaard:
@igormukhingmailcom, yes, that is definitely something for the next iteration of this feature :)

src/Resources/config/doctrine/model/GiftCardConfiguration.orm.xml
```diff
@@ -0,0 +1,29 @@
...
+        </field>
+
+        <field name="updatedAt" column="updated_at" type="datetime" nullable="true">
+            <gedmo:timestampable on="update"/>
+        </field>
+
+        <one-to-many field="channels" target-entity="Sylius\Component\Channel\Model\ChannelInterface" mapped-by="giftCardConfiguration">
+        </one-to-many>
+
+        <one-to-many field="images" target-entity="Setono\SyliusGiftCardPlugin\Model\GiftCardConfigurationImageInterface"
```
loevgaard: Why do we need multiple images here?

src/Model/ChannelTrait.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Setono\SyliusGiftCardPlugin\Model;
+
+use Doctrine\ORM\Mapping as ORM;
+
+trait ChannelTrait
```
loevgaard: This means we have to ask the user of the plugin to change his channel mapping also. I am not sure we need this, do we?

src/Resources/config/doctrine/model/GiftCardConfiguration.orm.xml
```diff
@@ -0,0 +1,29 @@
...
+        <field name="enabled" type="boolean"/>
+
+        <field name="createdAt" column="created_at" type="datetime">
+            <gedmo:timestampable on="create"/>
+        </field>
+
+        <field name="updatedAt" column="updated_at" type="datetime" nullable="true">
+            <gedmo:timestampable on="update"/>
+        </field>
+
```
loevgaard: You are missing a relation to a locale, aren't you?

src/Resources/config/doctrine/model/GiftCardConfiguration.orm.xml
Roshyo: This is to anticipate potential other images without having to rewrite everything.

src/Model/ChannelTrait.php
Roshyo: About the relation, 1 channel may have 1 configuration, and 1 configuration may be used by multiple channels. So I believe the owner has to be the channel, hasn't it ?

src/Resources/config/doctrine/model/GiftCardConfiguration.orm.xml
Roshyo: Yes indeed, will be fixed.

src/Model/ChannelTrait.php
Roshyo: Oh, but with the Locale interfering in relation, indeed the relation will change a bit

loevgaard: > About the relation, 1 channel may have 1 configuration, and 1 configuration may be used by multiple channels. So I believe the owner has to be the channel, hasn't it ?

Yes, that is correct. Do you think we should use a join table instead?

Roshyo: I believe we require a full entity to make the join. It will contain `Channel`, `LocaleCode` and `Configuration`

src/Model/GiftCardConfiguration.php
```diff
@@ -110,7 +110,7 @@ public function setBackgroundImage(?GiftCardConfigurationImageInterface $image):
     }
 
     /**
-     * {@inheritDoc}
+     * {@inheritdoc}
```
igormukhingmailcom: We have kinda convention to not use phpdocs like `inheritdoc`.

src/Controller/Action/DownloadGiftCardPdfAction.php
```diff
@@ -25,18 +19,18 @@ final class DownloadGiftCardPdfAction
...
-        }
-        if (!$this->authChecker->isGranted(GiftCardVoter::READ, $giftCard)) {
-            $this->flashBag->add('error', 'setono_sylius_gift_card.gift_card.read_error');
-
-            return new RedirectResponse(filter_var($request->headers->get('referer'), \FILTER_SANITIZE_URL));
-        }
-
-        $configuration = $giftCard->getConfiguration();
-    }
+//    public function __invoke(Request $request, int $giftCardId): Response
```
igormukhingmailcom: Please, do not commit commented code blocks (even in intermediate commits)

Roshyo: Ok

src/Model/GiftCardConfiguration.php
Roshyo: Shall I just remove the whole docblock ?

loevgaard: Yes :)

src/Controller/Action/DownloadGiftCardPdfAction.php
```diff
@@ -0,0 +1,90 @@
...
+    {
+        $giftCard = $this->giftCardRepository->find($id);
+        if (!$giftCard instanceof GiftCardInterface) {
+            throw new NotFoundHttpException('Gift card not found');
+        }
+        if (!$this->authChecker->isGranted([GiftCardVoter::READ], $giftCard)) {
+            $this->flashBag->add('error', 'setono_sylius_gift_card.gift_card.read_error');
+
+            /** @var string $redirectUrl */
+            $redirectUrl = filter_var($request->headers->get('referer'), \FILTER_SANITIZE_URL);
```
loevgaard: Any reason to sanitize this?

src/EventListener/GiftCardCreationSubscriber.php
```diff
@@ -0,0 +1,44 @@
...
+
+        if (!$giftCard instanceof GiftCardInterface) {
+            throw new UnexpectedTypeException(
+                $giftCard,
+                GiftCardInterface::class
+            );
+        }
+
+        $configuration = $this->configurationProvider->getConfigurationForGiftCard($giftCard);
+        $giftCard->setConfiguration($configuration);
```
loevgaard: Do we need to associate a gift card with a configuration? It only applies when we want to download a PDF, right?

src/Form/Type/GiftCardConfigurationImageType.php
```diff
@@ -0,0 +1,34 @@
...
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+final class GiftCardConfigurationImageType extends ImageType
+{
+    public function buildForm(FormBuilderInterface $builder, array $options): void
+    {
+        parent::buildForm($builder, $options);
+
+        if ($options['remove_type']) {
+            $builder->remove('type');
```
loevgaard: What is it that this does? :)

src/Model/GiftCardChannelConfiguration.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        $this->locale = $locale;
+    }
+
+    public function getConfiguration(): ?GiftCardConfigurationInterface
+    {
+        return $this->configuration;
+    }
+
+    public function setConfiguration(?GiftCardConfigurationInterface $configuration): void
```
loevgaard:
```suggestion
    public function setConfiguration(GiftCardConfigurationInterface $configuration): void
```

src/Model/GiftCardChannelConfiguration.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        return $this->id;
+    }
+
+    public function getChannel(): ?SyliusChannelInterface
+    {
+        return $this->channel;
+    }
+
+    public function setChannel(?SyliusChannelInterface $channel): void
```
loevgaard:
```suggestion
    public function setChannel(SyliusChannelInterface $channel): void
```

src/Model/GiftCardChannelConfiguration.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        $this->channel = $channel;
+    }
+
+    public function getLocale(): ?LocaleInterface
+    {
+        return $this->locale;
+    }
+
+    public function setLocale(?LocaleInterface $locale): void
```
loevgaard:
```suggestion
    public function setLocale(LocaleInterface $locale): void
```

src/Resources/config/app/config.yaml
```diff
@@ -18,3 +19,9 @@ jms_serializer:
             setono-sylius-gift-card-sylius:
                 namespace_prefix: "Setono\\SyliusGiftCardPlugin\\Model"
                 path: "@SetonoSyliusGiftCardPlugin/Resources/config/serializer"
+
+liip_imagine:
+    filter_sets:
+        gift_card_pdf:
```
loevgaard:
```suggestion
        setono_sylius_gift_card_background:
```

To be consistent with Sylius I think the namespace should be included. I also renamed to `background`. Wdyt?

src/Controller/Action/DownloadGiftCardPdfAction.php
Roshyo: Not sanitizing this opens up to XSS leak

src/EventListener/GiftCardCreationSubscriber.php
Roshyo: Yes, but for immutable concept, I believe we should attach them.
A customer should expect to get his gift card the same way it was even if we change the image for new cards. WDYT ?

src/Form/Type/GiftCardConfigurationImageType.php
Roshyo: It is to allow having the `backgroundImage` field in `GiftCardConfigurationType`  instead of collection of images where you have to enter the type

src/Model/GiftCardChannelConfiguration.php
Roshyo: This cannot be done since SyliusInterface does not allow it

src/Resources/config/app/config.yaml
Roshyo: Yes, good point

src/Model/GiftCardChannelConfiguration.php
loevgaard: Oh yeah. Assert::notNull then I guess

src/EventListener/GiftCardCreationSubscriber.php
loevgaard: I agree, but it's still not immutable that way. Because the configuration is a reference I guess? :)

Roshyo: Good point.
Then ideed the relation here is just a 'shortcut' and is not really needed. We can only rely on the provider to have the information

src/Form/Type/GiftCardConfigurationImageType.php
```diff
@@ -0,0 +1,34 @@
...
+        }
+    }
+
+    public function configureOptions(OptionsResolver $resolver): void
+    {
+        parent::configureOptions($resolver);
+
+        $resolver->setDefault('remove_type', false);
+        $resolver->setAllowedTypes('remove_type', ['null', 'bool']);
```
loevgaard:
```suggestion
        $resolver->setAllowedTypes('remove_type', 'bool');
```

src/Model/GiftCardConfiguration.php
```diff
@@ -0,0 +1,162 @@
...
+        $this->images = new ArrayCollection();
+        $this->channelConfigurations = new ArrayCollection();
+    }
+
+    public function getId(): ?int
+    {
+        return $this->id;
+    }
+
+    public function getCode(): ?string
```
loevgaard: Maybe implement the `CodeAwareInterface` and I guess there is a trait also

src/Model/GiftCardConfiguration.php
```diff
@@ -0,0 +1,162 @@
...
+    /** @var string|null */
+    protected $code;
+
+    /** @var ImageInterface[]|Collection */
+    protected $images;
+
+    /** @var GiftCardChannelConfigurationInterface[]|Collection */
+    protected $channelConfigurations;
+
+    /** @var bool|null */
```
loevgaard:
```suggestion
    /** @var bool */
```

src/Model/GiftCardConfiguration.php
```diff
@@ -0,0 +1,162 @@
...
+            $this->channelConfigurations->removeElement($channelConfiguration);
+        }
+    }
+
+    public function isDefault(): ?bool
+    {
+        return $this->default;
+    }
+
+    public function setDefault(?bool $default): void
```
loevgaard:
```suggestion
    public function setDefault(bool $default): void
```

src/Model/GiftCardConfiguration.php
```diff
@@ -0,0 +1,162 @@
...
+
+    public function removeChannelConfiguration(GiftCardChannelConfigurationInterface $channelConfiguration): void
+    {
+        if ($this->hasChannelConfiguration($channelConfiguration)) {
+            $channelConfiguration->setConfiguration(null);
+            $this->channelConfigurations->removeElement($channelConfiguration);
+        }
+    }
+
+    public function isDefault(): ?bool
```
loevgaard:
```suggestion
    public function isDefault(): bool
```

src/Model/GiftCardConfiguration.php
Roshyo: The interface is implemented (in GiftCardConfigurationInterface) but there is no such trait...

igormukhingmailcom: Btw, missed useful traits like this should be added to SyliusCorePlugin.
But from other side, they can be used only at apps. Having SyliusCorePlugin as dependency at other plugins is not good.

PR #62, opened 2020-03-27
Match default generated file structures

TheDevilOnLine:
The namespaces/directories for Order and Product were different from the default generated files with a clean 1.7 installation, updated the README.md to reflect the default structure.


loevgaard:
Thank you, @TheDevilOnLine :)

PR #68, opened 2020-05-26
French translation

JulienLoison:
French translation, checked most of it but some parts may be too literal.


loevgaard:
Thank you, @JulienLoison 🎉

loevgaard:
Great profile picture :D

PR #72, opened 2020-07-20
fix(ajax): call ajax take time with big db. Limit to 10 should be good enough

chadyred:
**WHY ?**

With lots of email, it is slow to load data.

**HOW ?**

Add a parameter with limit to 10. For autocomplete, it is good enough.

### DIFF
--- a/src/Doctrine/ORM/CustomerRepositoryTrait.php
+++ b/src/Doctrine/ORM/CustomerRepositoryTrait.php
@@ -16,11 +16,12 @@ trait CustomerRepositoryTrait
      */
     abstract public function createQueryBuilder($alias, $indexBy = null);
 
-    public function findByEmailPart(string $email): array
+    public function findByEmailPart(string $email, int $limit = 10): array
     {
         return $this->createQueryBuilder('o')
             ->andWhere('o.email LIKE :phrase')
             ->setParameter('phrase', '%' . $email . '%')
+            ->setMaxResult($limit)
             ->getQuery()
             ->getResult()
         ;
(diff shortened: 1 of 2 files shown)


chadyred:
@loevgaard CI not pass for other reason.

loevgaard:
Hi @chadyred 

Thanks for your PR. Could you just rebase your branch onto master? I fixed the build

src/Doctrine/ORM/CustomerRepositoryTrait.php
```diff
@@ -16,11 +16,12 @@ trait CustomerRepositoryTrait
      */
     abstract public function createQueryBuilder($alias, $indexBy = null);
 
-    public function findByEmailPart(string $email): array
+    public function findByEmailPart(string $email, int $limit): array
```
loevgaard, line 19:
```suggestion
    public function findByEmailPart(string $email, int $limit = 10): array
```

src/Repository/CustomerRepositoryInterface.php
```diff
@@ -8,5 +8,5 @@
 
 interface CustomerRepositoryInterface extends BaseCustomerRepositoryInterface
 {
-    public function findByEmailPart(string $email): array;
+    public function findByEmailPart(string $email, int $limit): array;
```
loevgaard, line 11:
```suggestion
    public function findByEmailPart(string $email, int $limit = 10): array;
```

src/Resources/config/routes/admin_ajax.yaml
```diff
@@ -11,6 +11,7 @@ setono_sylius_gift_card_admin_ajax_customer_by_email_phrase:
                 method: findByEmailPart
                 arguments:
                     email: $phrase
+                    limit: 10
```
loevgaard, line 14: You can revert this change :)

chadyred:
Done !

loevgaard:
Released as 0.8.5 :) 

Thank you!

PR #89, opened 2020-08-06
Handle the empty background image

### LINKED ISSUE #66: Exception when configuration doesn't have background image
Here is an image of my gift card configuration

image

Then I go to the index of gift cards and press `PDF`. I am then presented with this exception:

image

Roshyo:
Also make it required
Fixes #66

### DIFF
--- a/src/Form/Type/GiftCardConfigurationType.php
+++ b/src/Form/Type/GiftCardConfigurationType.php
@@ -9,6 +9,8 @@ use Symfony\Component\Form\Extension\Core\Type\CheckboxType;
 use Symfony\Component\Form\Extension\Core\Type\CollectionType;
 use Symfony\Component\Form\Extension\Core\Type\TextType;
 use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\Form\FormInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
 
 final class GiftCardConfigurationType extends AbstractResourceType
 {
@@ -42,5 +44,19 @@ final class GiftCardConfigurationType extends AbstractResourceType
     public function getBlockPrefix(): string
     {
         return 'setono_sylius_gift_card_gift_card_configuration';
+    }
+
+    public function configureOptions(OptionsResolver $resolver): void
+    {
+        parent::configureOptions($resolver);
+        $resolver->setDefault('validation_groups', function (FormInterface $form): array {
+            $validationGroups = $this->validationGroups;
+            $data = $form->getData();
+            if (null === $data->getId()) {
+                $validationGroups[] = 'setono_sylius_gift_card_create';
+            }
+
+            return $validationGroups;
+        });
     }
 }
--- a/src/Form/Type/GiftCardConfigurationType.php
+++ b/src/Form/Type/GiftCardConfigurationType.php
@@ -9,8 +9,6 @@ use Symfony\Component\Form\Extension\Core\Type\CheckboxType;
 use Symfony\Component\Form\Extension\Core\Type\CollectionType;
 use Symfony\Component\Form\Extension\Core\Type\TextType;
 use Symfony\Component\Form\FormBuilderInterface;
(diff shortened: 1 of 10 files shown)


src/Form/Type/GiftCardConfigurationType.php
```diff
@@ -43,4 +45,18 @@ public function getBlockPrefix(): string
...
+        $resolver->setDefault('validation_groups', function (FormInterface $form): array {
+            $validationGroups = $this->validationGroups;
+            $data = $form->getData();
+            if (null === $data->getId()) {
+                $validationGroups[] = 'setono_sylius_gift_card_create';
+            }
+
+            return $validationGroups;
+        });
+    }
```
loevgaard, line 61: I don't understand why this is needed?

Roshyo: Oh, it is not anymore. It was on my first attempt. I will remove that

src/Resources/config/services/validator.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
loevgaard, line 1: Is this file included in the root `services.xml`?

src/Resources/config/validation/GiftCardConfiguration.xml
```diff
@@ -9,6 +9,13 @@
             <option name="fields">code</option>
             <option name="groups">setono_sylius_gift_card</option>
         </constraint>
+        <constraint name="Setono\SyliusGiftCardPlugin\Validator\Constraints\HasBackgroundImage">
+            <option name="groups">
+                <value>setono_sylius_gift_card</value>
+                <value>setono_sylius_gift_card_create</value>
```
loevgaard, line 15: Why do we need the `setono_sylius_gift_card_create` here?

src/Resources/translations/validators.en.yml
```diff
@@ -15,6 +15,9 @@ setono_sylius_gift_card:
             integer: Amount must be an integer  
             min: Amount cannot be lower than 1
             greater_than_or_equal: The amount must be greater than or equal to 1
+    gift_card_configuration:
+        create:
```
loevgaard, line 19: Why this sub namespace?

src/Validator/Constraints/HasBackgroundImageValidator.php
```diff
@@ -0,0 +1,31 @@
...
+        if (!$constraint instanceof HasBackgroundImage) {
+            throw new UnexpectedTypeException($constraint, HasBackgroundImage::class);
+        }
+
+        if (!$value instanceof GiftCardConfigurationInterface) {
+            return;
+        }
+
+        $backgroundImage = $value->getBackgroundImage();
+        if (!$backgroundImage instanceof GiftCardConfigurationImageInterface || null === $backgroundImage->getPath()) {
```
loevgaard, line 26:
```suggestion
        if (null === $backgroundImage || null === $backgroundImage->getPath()) {
```

src/Resources/config/services/validator.xml
Roshyo: No it is not, but for some reasons it works fine.
I'm adding it

src/Resources/translations/validators.en.yml
Roshyo: Not required anymore

src/Resources/config/validation/GiftCardConfiguration.xml
Roshyo: We don't anymore

src/Validator/Constraints/HasBackgroundImageValidator.php
Roshyo: Is there a specific reason for that ? Or just something like line length ?
I tend to prefer the `instanceof` because it then gives a clear idea of what we expect. (Side effect is autocompletion in IDE as well)

PR #91, opened 2020-08-07
Allowing to change the redirect target of add/remove gift card in cart

### LINKED ISSUE #69: Allow to change the redirect URL when adding or deleting gift cards
I had to allow the customer to add and remove gift card on the payment step of the checkout process. I needed the customer to be redirected to the payment step and not to the cart summary page. 
I had to override `AddGiftCardToOrderAction`, `RemoveGiftCardFromOrderAction` and `addGiftCardToOrder.html.twig` to achieve that (replacing `sylius_shop_cart_summary` by the route I wanted to redirect to)

May be it could be done with the referrer from the request, this way no matter where your user decides to implement the plugin, the customer will be redirected to the page the request came from ?

Roshyo:
Fixes #69

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,3 +281,5 @@ setono_sylius_gift_card_shop_remove_gift_card_from_order:
 ```
 
 The same applies for the `setono_sylius_gift_card_shop_partial_add_gift_card_to_order` route
+
+You can also override or decorate the service `setono_sylius_gift_card.resolver.redirect_url` to define a more custom way of redirecting
--- a/README.md
+++ b/README.md
@@ -262,3 +262,22 @@ Please, run `composer all` to run all checks and tests before making pull reques
 [link-packagist]: https://packagist.org/packages/setono/sylius-gift-card-plugin
 [link-github-actions]: https://github.com/Setono/SyliusGiftCardPlugin/actions
 [link-code-quality]: https://scrutinizer-ci.com/g/Setono/SyliusGiftCardPlugin
+
+# Configuration
+
+## Change redirect routes on add/remove gift card to/from order
+
+You can customize where you will be redirected after adding or removing a gift card. To do so, you can simply change the route configuration :
+
+```yaml
+setono_sylius_gift_card_shop_remove_gift_card_from_order:
+    path: /gift-card/{giftCard}/remove-from-order
+    methods: GET
+    defaults:
+        _controller: setono_sylius_gift_card.controller.action.remove_gift_card_from_order
+        redirect:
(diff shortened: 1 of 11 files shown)


src/Resolver/RedirectRouteResolver.php
```diff
@@ -0,0 +1,51 @@
...
+namespace Setono\SyliusGiftCardPlugin\Resolver;
+
+use function array_key_exists;
+use Exception;
+use const FILTER_SANITIZE_URL;
+use function filter_var;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+
+final class RedirectRouteResolver implements RedirectRouteResolverInterface
```
loevgaard, line 14: The name of this is wrong. You are not resolving a route, you are resolving an URL

src/Resolver/RedirectRouteResolver.php
```diff
@@ -0,0 +1,51 @@
...
+            $redirect = $request->attributes->get('redirect');
+
+            try {
+                if (array_key_exists(0, $redirect)) {
+                    return $this->router->generate($redirect[0]);
+                }
+                if (array_key_exists('route', $redirect) && array_key_exists('parameters', $redirect)) {
+                    return $this->router->generate($redirect['route'], $redirect['parameters']);
+                }
+            } catch (Exception $exception) {
```
loevgaard, line 36: This exception is throw is the route doesn't exist, right? If yes, I don't think we should catch it. For now I think it's better to surface this error, so a developer can fix the error

README.md
```diff
@@ -262,3 +262,22 @@ Please, run `composer all` to run all checks and tests before making pull reques
 [link-packagist]: https://packagist.org/packages/setono/sylius-gift-card-plugin
 [link-github-actions]: https://github.com/Setono/SyliusGiftCardPlugin/actions
 [link-code-quality]: https://scrutinizer-ci.com/g/Setono/SyliusGiftCardPlugin
+
+# Configuration
+
+## Change redirect routes on add/remove gift card to/from order
+
+You can customize where you will be redirected after adding or removing a gift card. To do so, you can simply change the route configuration :
```
loevgaard, line 270: ... or override the service for the url resolver

src/Resolver/RedirectRouteResolver.php
Roshyo: Yes, that's right. Ok, it makes sense. I thought it would be more permissive this way, but that just sounds useless and it can only raise weird behaviors

PR #92, opened 2020-08-08
Allowing to use the same input for PromotionCoupon and GiftCards

### LINKED ISSUE #59: Allow to enter GiftCard code in the same input as coupon
It would be useful to allow users to enter the Gift card code in the same input as regular coupon. It should also be a possibility, not the unique behavior. So we could add in admin panel a new configuration for GiftCard that would be like "Have 2 separate inputs in cart ?" 

If yes : keep behavior like now
If no : Hook on "coupon application" and check whether it's a coupon (default behavior) or a Gift Card code.

We would also need to add a validator to GiftCardCode so they are never the same as coupon

Roshyo:
Fixes #59

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,8 @@ $bundles = [
 
 You will find the templates you need to override in the [test application](https://github.com/Setono/SyliusGiftCardPlugin/tree/master/tests/Application/templates).
 
+If you enable the same input for Promotion and GiftCard (see [here](#using-same-input-for-promotion-and-giftcard)), remove all content of `SyliusShopBundle/Cart/Summary/_coupon.html.twig` since it is not needed.
+
 ### Extend `Product`, `Order`, `OrderRepository`, and `CustomerRepository`
 
 **Extend `Product`**
@@ -283,3 +285,14 @@ setono_sylius_gift_card_shop_remove_gift_card_from_order:
 The same applies for the `setono_sylius_gift_card_shop_partial_add_gift_card_to_order` route
 
 You can also override or decorate the service `setono_sylius_gift_card.resolver.redirect_url` to define a more custom way of redirecting
+
+## Using same input for Promotion and GiftCard
+
+You can configure the shop to use same input for GiftCards and Promotion coupons. To do so, set this configuration :
+
+```yaml
+# config/packages/setono_sylius_gift_card.yaml
+setono_sylius_gift_card:
+    cart:
+        use_same_input_for_promotion_and_gift_card: false
+```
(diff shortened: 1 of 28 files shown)


README.md
```diff
@@ -283,3 +285,14 @@ setono_sylius_gift_card_shop_remove_gift_card_from_order:
...
+
+## Using same input for Promotion and GiftCard
+
+You can configure the shop to use same input for GiftCards and Promotion coupons. To do so, set this configuration :
+
+```yaml
+# config/packages/setono_sylius_gift_card.yaml
+setono_sylius_gift_card:
+    cart:
+        use_same_input_for_promotion_and_gift_card: false
```
loevgaard, line 297:
```suggestion
        use_same_input_for_promotion_and_gift_card: true
```

src/Applicator/CouponCodeApplicator.php
```diff
@@ -0,0 +1,57 @@
...
+use Doctrine\ORM\EntityManagerInterface;
+use Setono\SyliusGiftCardPlugin\Model\OrderInterface;
+use Sylius\Component\Core\Model\PromotionCouponInterface;
+use Sylius\Component\Order\Processor\OrderProcessorInterface;
+use Sylius\Component\Promotion\Action\PromotionApplicatorInterface;
+use Sylius\Component\Promotion\Repository\PromotionCouponRepositoryInterface;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+use Webmozart\Assert\Assert;
+
+final class CouponCodeApplicator implements CouponCodeApplicatorInterface
```
loevgaard, line 16: Doesn't Sylius have a class that does this?

loevgaard (changes requested):
I think you should come up with a solution where you use different classes for the two scenarios (with and without using same input for coupon and gift card). I think this will make it much cleaner.

This means a form for both scenarios, a controller etc. Right now I think the controller for example is too complex

src/Applicator/CouponCodeApplicator.php
Roshyo: Sylius has applicators and processors, but none of them is flushing. So this applicator is just there to simplify the dependencies and to have a simpler code (controller especially)

demo18:
hello, could it be merged?

jeandanyel:
> could it be merged

it would be nice :)

Roshyo:
Hey there. There is still some work to do on this matter to simplify things and make it easier to read/extend.

But this issue is in my TODO and it should be dealt with in the upcoming weeks.

Issue #98, opened 2020-09-08
The background image is required


sabrinefk13:
![gift-card-conf](https://user-images.githubusercontent.com/57678173/92441918-540e1000-f1af-11ea-9b0d-8009de1417ef.png)

I cannot create a Gift card configurations. I have a back error **The background image is required**. This is caused by **HasBackgroundImageValidator.php** We must test if the file in the BackgroundImage (**$backgroundImage->getFile()->getPath()**) is not empty and not if **$backgroundImage->getPath()** is empty.

PR #99, opened 2020-09-08
Fix HasBackgroundImageValidator

sabrinefk13:
The background image is required (https://github.com/Setono/SyliusGiftCardPlugin/issues/98)


Roshyo:
Hi, and thank you for your PR :) 

There is just an issue that would be raised in the PDF : 

```
background: transparent url("{{ path }}") no-repeat left bottom;
```
And `path` would not be defined. SO we should just do a :

```
{% set path = '' %}
```
in the beginning of the template to fix that.

Can you take care of that @sabrinefk13 ?

sabrinefk13:
Hi, @Roshyo 
I have already tested to put the 
{% set path = '' %}
but it does not work :(

Roshyo:
What do you mean ?
The problem that I see here is that with your change on the validator, the backgroundImage can be null when generating the PDF.
If so, the variable `path` won't be defined in Twig, and when it will be called, it will simply crash. Won't it ?
So we need to make sure it is at least defined : 

```
{% set path = '' %}
{% if configuration.backgroundImage is not empty %}
    {% set path = configuration.backgroundImage.path|imagine_filter('setono_sylius_gift_card_background') %}
{% endif %}
```

Roshyo (approved):

Roshyo:
Thank you @sabrinefk13 !

loevgaard:
Released in `v0.10.1` :)

PR #115, opened 2020-12-03
Fixed: Setters

igormukhingmailcom:


loevgaard (approved):
Almost an infinite loop :D

PR #124, opened 2021-02-12
Add filters

igormukhingmailcom:


src/Doctrine/ORM/GiftCardRepository.php
```diff
@@ -12,11 +12,14 @@
 use Sylius\Component\Core\Model\CustomerInterface;
 use Sylius\Component\Core\Model\OrderItemUnitInterface;
 
-final class GiftCardRepository extends EntityRepository implements GiftCardRepositoryInterface
+/** not final */ class GiftCardRepository extends EntityRepository implements GiftCardRepositoryInterface
```
loevgaard, line 15: This was an error in the first place. Just remove this comment. We won't let it happen again :D

src/Grid/FieldTypes/StringFieldType.php
```diff
@@ -0,0 +1,37 @@
...
+        try {
+            $value = $this->propertyAccessor->getValue($data, $field->getPath());
+        } catch (UnexpectedTypeException $e) {
+            return '';
+        }
+
+        return htmlspecialchars((string) $value);
+    }
+
+    public function configureOptions(OptionsResolver $resolver): void
```
loevgaard, line 34: Is this empty method needed / is it in the interface?

src/Resources/config/services/field_types.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+
+<container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+    <services>
+        <defaults public="true" />
```
loevgaard, line 5: Should this be public?

loevgaard (changes requested):

loevgaard (approved):

src/Grid/FieldTypes/StringFieldType.php
igormukhingmailcom: Yes, it in interface

src/Resources/config/services/field_types.xml
igormukhingmailcom: In original `vendor/sylius/grid-bundle/src/Bundle/Resources/config/services/field_types.xml` we have `<defaults public="true" />` too, but I've tried and looks like it works without that

PR #128, opened 2021-03-22
Replace deprecated view handler with twig rendering

lruozzi9:
HI!
As reported in the [FOSRestBundle/UPGRADING-3.0.md](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/3.x/UPGRADING-3.0.md) the method setTemplate has been removed from the View/View class.
To resolve the issue this PR will replace the unnecessary `fos_rest.view_handler.default` service with the `twig` service.

### DIFF
--- a/src/Controller/Action/AddGiftCardToOrderAction.php
+++ b/src/Controller/Action/AddGiftCardToOrderAction.php
@@ -4,8 +4,6 @@ declare(strict_types=1);
 
 namespace Setono\SyliusGiftCardPlugin\Controller\Action;
 
-use FOS\RestBundle\View\View;
-use FOS\RestBundle\View\ViewHandlerInterface;
 use Setono\SyliusGiftCardPlugin\Applicator\GiftCardApplicatorInterface;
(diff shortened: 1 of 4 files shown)


loevgaard:
Will this work in Sylius 1.7, 1.8?

lruozzi9:
Hi @loevgaard! Well, after some research, I found that Sylius 1.7 use a release subsequent the [2.12](https://github.com/Sylius/Sylius/blob/v1.7.0/composer.json) and in Twig v2.12 the class Environment is [present](https://github.com/twigphp/Twig/blob/v2.12.0/src/Environment.php). So, I haven't tested it, but I'm pretty sure it should works!

lruozzi9:
Apart from the fact that the setTemplate method was deprecated and removed, I think this was the right solution right away.
I understand that FOSRestBundle is specific for rendering serialized data for REST API and not for rendering views.

loevgaard:
Hi, @lruozzi9 

Could you rebase this PR and see if it works? The build works on the master branch now

lruozzi9:
Ok

Roshyo:
This can be closed IMO as it is part of job done in #139

PR #132, opened 2021-03-29
Adjust template reference

lruozzi9:
As shown [here](https://docs.sylius.com/en/1.8/cookbook/entities/custom-model.html#define-routing-for-entity-administration) Sylius use the double backslash to refer to the template in the controller. This also solves a problem using Symfony 5.2.


loevgaard:
This was fixed in the last major upgrade. Sorry for not seeing it here first :|

lruozzi9:
No problem 😉. I needed it urgently so I needed a patch for my project.

PR #133, opened 2021-03-29
Use Entity Manager Interface instead of Object Manager

lruozzi9:
Hello! Using the combination of Sylius 1.9, Symfony 5.2 and Doctrine 2 there is some error generated by the incompatibility between the sylius.manager.order service and the ObjectManager class. Using EntityManagerInterface to define the property will solve the problem.


loevgaard:
This is not the way to solve it, but it will be solved in this PR: https://github.com/Setono/SyliusGiftCardPlugin/pull/136. The namespace should just be changed.

lruozzi9:
Ok @loevgaard, I'll watch your changes.

PR #139, opened 2021-08-18
Allow to use Sylius from 1.7 to 1.10 and php 7.4 and 8.0

Roshyo:


composer.json
```diff
@@ -4,14 +4,13 @@
     "description": "Gift card plugin for Sylius",
     "license": "MIT",
     "require": {
-        "php": ">=7.4",
+        "php": ">=7.4 || ^8.0",
```
loevgaard: Doesn't make sense to add ` || ^8.0` when the current is `>=7.4` :)

composer.json
```diff
@@ -31,19 +30,24 @@
...
         "phpspec/phpspec": "^7.0",
         "phpunit/phpunit": "^9.5",
         "psalm/plugin-phpunit": "^0.16.0",
         "psalm/plugin-symfony": "^2.1",
-        "roave/security-advisories": "dev-latest",
         "setono/code-quality-pack": "^2.1",
-        "setono/sylius-behat-pack": "^0.1",
-        "setono/sylius-catalog-promotion-plugin": "^0.2.0",
-        "sylius/sylius": "~1.7.0",
+        "setono/job-status-bundle": "^0.2.1",
```
loevgaard: Shouldn't be added here. Isn't this already added in the catalog promotion plugin?

composer.json
```diff
@@ -31,19 +30,24 @@
...
+        "setono/sylius-behat-pack": "^0.1 || ^0.2",
+        "setono/sylius-catalog-promotion-plugin": "^0.3 || ^0.4",
```
loevgaard: Dev deps should normally not be widened, but just be increased, i.e. `^0.2` and `^0.4` respectively :D

src/DependencyInjection/SetonoSyliusGiftCardExtension.php
```diff
@@ -17,6 +18,7 @@ public function load(array $configs, ContainerBuilder $container): void
         $config = $this->processConfiguration($this->getConfiguration([], $container), $configs);
         $loader = new XmlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
 
+        Assert::integer($config['code_length']);
```
loevgaard: Usually I just phpdoc these when in the `Extension` since they come from the `Configuration` which in turn is provided by Symfony :)

src/Resources/config/services/controller.xml
```diff
@@ -28,13 +28,13 @@
         <service id="setono_sylius_gift_card.controller.action.gift_card_balance"
                  class="Setono\SyliusGiftCardPlugin\Controller\Action\GiftCardBalanceAction">
             <argument type="service" id="setono_sylius_gift_card.repository.gift_card"/>
-            <argument type="service" id="fos_rest.view_handler.default"/>
+            <argument type="service" id="twig"/>
```
loevgaard, line 31: When we release we need to keep in mind that this is a BC break

tests/Application/composer.json
```diff
@@ -1,5 +1,8 @@
 {
     "name": "sylius/plugin-skeleton-test-application",
     "description": "Sylius application for plugin testing purposes (composer.json needed for project dir resolving)",
-    "license": "MIT"
+    "license": "MIT",
+    "require-dev": {
+        "symfony/apache-pack": "^1.0"
+    }
```
loevgaard: ?

tests/Application/config/sylius/1.10/bundles.php
```diff
@@ -0,0 +1,10 @@
+<?php
```
loevgaard: I understand the Sylius Plugin Skeleton does this, but the maintenance burden with this approach is just too high in my opinion, therefore I think we should just upgrade the application to 1.10 and that's it.

loevgaard (changes requested):

composer.json
```diff
@@ -31,19 +30,24 @@
         "twig/twig": "^2.14",
         "webmozart/assert": "^1.10"
     },
+    "conflict": {
+        "doctrine/doctrine-bundle": "2.3.0",
+        "sylius-labs/coding-standard": "4.1.0",
```
loevgaard: Require `setono/code-quality-pack:^2.1.2` instead

loevgaard (changes requested):

composer.json
Roshyo: What do you mean, should I rather add `>= 8.0` ?

tests/Application/config/sylius/1.10/bundles.php
Roshyo: But what would happen for older versions ? We won't support them anymore ?

I can understand for 1.7 (perhaps also 1.8) but 1.9 is still quite recent and not everyone updated yet.

WDYT ?

composer.json
```diff
@@ -31,19 +30,22 @@
...
     "require-dev": {
-        "doctrine/doctrine-bundle": "^1.2.13",
+        "friendsofsymfony/oauth-server-bundle": ">2.0.0-alpha.0 ^2.0@dev",
         "matthiasnoback/symfony-config-test": "^4.2",
         "matthiasnoback/symfony-dependency-injection-test": "^4.1",
         "phpspec/phpspec": "^7.0",
         "phpunit/phpunit": "^9.5",
         "psalm/plugin-phpunit": "^0.16.0",
         "psalm/plugin-symfony": "^2.1",
-        "roave/security-advisories": "dev-latest",
```
loevgaard, line 42: No reason to remove this I guess? :)

src/DependencyInjection/SetonoSyliusGiftCardExtension.php
```diff
@@ -17,7 +17,9 @@ public function load(array $configs, ContainerBuilder $container): void
         $config = $this->processConfiguration($this->getConfiguration([], $container), $configs);
         $loader = new XmlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
 
-        $container->setParameter('setono_sylius_gift_card.code_length', $config['code_length']);
+        /** @var int $codeLength */
```
loevgaard: Ah, I just do it on the `$config`. Something like:

```php
/** @var array{code_length: int} $config */
```

composer.json
```diff
@@ -31,19 +30,22 @@
         "twig/twig": "^2.14",
         "webmozart/assert": "^1.10"
     },
+    "conflict": {
+        "doctrine/doctrine-bundle": "2.3.0",
+        "symfony/doctrine-bridge": "4.4.16"
```
loevgaard: Do you remember what was wrong with these? Because I don't remember having problems with them on other plugins

src/DependencyInjection/SetonoSyliusGiftCardExtension.php
Roshyo: Doing it without setting it results in that: 

image

composer.json
Roshyo: No I don't. Probably at some version like 1.8/1.9. Now it seems to be okay without those two.

loevgaard (approved):

PR #140, opened 2021-08-23
Adding ApiPlatform integration to GiftCards

Roshyo:


src/Resources/config/api_resources/GiftCard.xml
```diff
@@ -0,0 +1,63 @@
+<?xml version="1.0" ?>
+
+<resources xmlns="https://api-platform.com/schema/metadata"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="https://api-platform.com/schema/metadata https://api-platform.com/schema/metadata/metadata-2.0.xsd"
+>
+    <resource class="%setono_sylius_gift_card.model.gift_card.class%" shortName="GiftCard">
```
loevgaard, line 7: > `shortName="GiftCard"`

Is this the convention for short names? What is it used for? Should it be namespaced?

tests/Application/config/sylius/1.10/packages/security.yaml
```diff
@@ -1,13 +1,14 @@
 parameters:
```
loevgaard: Shouldn't this file just be in `tests/Application/config`?

src/Resources/config/api_resources/GiftCard.xml
Roshyo: It seems to be the one from Sylius. It is used to display the resource name in API doc
image

It should not be namespaced imo. But it seems that by default the name will be `GiftCard` for us. So perhaps not really needed here

tests/Application/config/sylius/1.10/packages/security.yaml
Roshyo: Copied from Sylius skeleton. And this would fail in <=1.9, so it has to be in 1.10 config

composer-require-checker.json
```diff
@@ -15,8 +15,10 @@
     "string",
     "true",
     "void",
-    "Sylius\\Bundle\\CoreBundle\\Application\\SyliusPluginTrait",
     "Setono\\SyliusCatalogPromotionPlugin\\Rule\\Rule",
+    "Sylius\\Bundle\\ApiBundle\\Command\\CommandAwareDataTransformerInterface",
```
loevgaard, line 19: Isn't these files this: https://packagist.org/packages/sylius/api-bundle ? If so then we should add this as a direct dependency in `require`

tests/Api/Controller/Action/UploadGiftCardConfigurationImageActionTest.php
```diff
@@ -0,0 +1,285 @@
...
+        $uploadGiftCardConfigurationImageAction = new UploadGiftCardConfigurationImageAction(
+            $giftCardConfigurationImageFactory,
+            $giftCardConfigurationImageRepository,
+            $imageUploader,
+            $iriConverter
+        );
+        $uploadGiftCardConfigurationImageAction($request);
+    }
+
+    public function testDeleteOldImageOfSameType(): void
```
loevgaard: Maybe we should have a video call tomorrow about PHPUnit also, but I will just give a few of my thoughts here.

The 'new' standard is to name the methods as you do in phpspec, like 'it_deletes_old_image_of_same_type' and then annotate the method with:

```php
/**
 * @test
 */
```

Below here you should just use concrete classes and instantiate them. You will find this is much easier for the most part. Here is a good article about this: https://blog.frankdejonge.nl/testing-without-mocking-frameworks/ (first found on twitter ;)))

When you need to mock something I still like the way phpspec does it if it fits. For this you can use this library: https://github.com/phpspec/prophecy-phpunit

src/Resources/config/services/api/controller.xml
```diff
@@ -0,0 +1,14 @@
...
+
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+        <service id="setono_sylius_gift_card_plugin.api.upload_gift_card_configuration_image"
+                 class="Setono\SyliusGiftCardPlugin\Api\Controller\Action\UploadGiftCardConfigurationImageAction"
+                 autowire="true">
+            <tag name="controller.service_arguments" />
```
loevgaard, line 15: Isn't it enough to declare the service public?

src/Resources/config/services/api/controller.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+        <service id="setono_sylius_gift_card_plugin.api.upload_gift_card_configuration_image"
+                 class="Setono\SyliusGiftCardPlugin\Api\Controller\Action\UploadGiftCardConfigurationImageAction"
+                 autowire="true">
```
loevgaard: We should not use autoconfigure and autowire in plugins, only in applications

src/Api/Command/AssociateConfigurationToChannel.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Setono\SyliusGiftCardPlugin\Api\Command;
+
+class AssociateConfigurationToChannel implements ConfigurationCodeAwareInterface
+{
+    public ?string $configurationCode = null;
```
loevgaard, line 9: Why isn't this property instantiated in the constructor?

loevgaard (changes requested):

composer-require-checker.json
Roshyo: For some reasons (maybe my lack of knowledge with composer) but I don't see this package anywhere in installed list. And adding it as a dependency has 0 impact neither

src/Resources/config/services/api/controller.xml
Roshyo: We still need to tag it as a Controller, don't we ? So services are injected into methods

src/Api/Command/AssociateConfigurationToChannel.php
Roshyo: Because this information is not coming from PostData as are the other ones, but it is coming from ApiPlatform (giving to URL as object to populate)

composer-require-checker.json
loevgaard: Okay, that's how it is with some of Sylius' packages. I didn't know that was also the case with the API bundle.

src/Resources/config/services/api/controller.xml
loevgaard: I am pretty sure you are right here

phpunit.xml.dist
```diff
@@ -11,7 +11,7 @@
     </coverage>
     <testsuites>
         <testsuite name="SetonoSyliusGiftCardPlugin Test Suite">
-            <directory>tests</directory>
+            <directory>tests/PHPUnit</directory>
```
loevgaard:
```suggestion
            <directory>tests/Unit</directory>
```

Roshyo: Ok, I thought of `PHPUnit` instead because this folder can contains other kind of tests. But ok, I will rename it as it is often named that way

tests/PHPUnit/Api/CommandHandler/AssociateConfigurationToChannelHandlerTest.php
```diff
@@ -0,0 +1,250 @@
...
+        $configuration = new GiftCardConfiguration();
+        $channel = new Channel();
+        $locale = new Locale();
+        $giftCardChannelConfiguration = new GiftCardChannelConfiguration();
+
+        $giftCardConfigurationRepository = $this->prophesize(RepositoryInterface::class);
+        $channelRepository = $this->prophesize(ChannelRepositoryInterface::class);
+        $localeRepository = $this->prophesize(RepositoryInterface::class);
+        $giftCardChannelConfigurationRepository = $this->prophesize(RepositoryInterface::class);
+        $giftCardChannelConfigurationFactory = $this->prophesize(FactoryInterface::class);
```
loevgaard: Actually you don't need to mock the factory either, but it's okay :)

tests/PHPUnit/Api/CommandHandler/AssociateConfigurationToChannelHandlerTest.php
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    /**
+     * @test
+     */
+    public function it_does_nothing_if_association_already_exists(): void
+    {
+        $configuration = $this->createMock(GiftCardConfiguration::class);
+        $channel = $this->createMock(Channel::class);
+        $locale = $this->createMock(Locale::class);
```
loevgaard: Leftover mocks :D

loevgaard (approved):

PR #143, opened 2021-08-27
Expose Admin resend email route

Roshyo:


src/Api/Controller/Action/ResendGiftCardEmailAction.php
```diff
@@ -0,0 +1,36 @@
...
+final class ResendGiftCardEmailAction
+{
+    private GiftCardEmailManagerInterface $giftCardEmailManager;
+
+    public function __construct(GiftCardEmailManagerInterface $giftCardEmailManager)
+    {
+        $this->giftCardEmailManager = $giftCardEmailManager;
+    }
+
+    public function __invoke(GiftCardInterface $data): Response
```
loevgaard, line 21:
```suggestion
    public function __invoke(GiftCardInterface $giftCard): Response
```

src/Api/Controller/Action/ResendGiftCardEmailAction.php
```diff
@@ -0,0 +1,36 @@
...
+    {
+        $giftCard = $data;
```
loevgaard:
```suggestion
```

src/Api/Controller/Action/ResendGiftCardEmailAction.php
```diff
@@ -0,0 +1,36 @@
...
+        if (($order = $giftCard->getOrder()) instanceof OrderInterface) {
```
loevgaard:
```suggestion
        if (($order = $giftCard->getOrder()) !== null) {
```

I suppose this is the same because the `OrderInterface` is enforced by type hints, right? The same goes for the code below

src/Resources/config/api_resources/GiftCard.xml
```diff
@@ -40,6 +40,15 @@
                 </attribute>
             </itemOperation>
 
+            <itemOperation name="admin_resend_email">
+                <attribute name="method">GET</attribute>
+                <attribute name="path">/admin/gift-cards/{code}/resend-email</attribute>
+                <attribute name="controller">setono_sylius_gift_card_plugin.api.resend_gift_card_email</attribute>
+                <attribute name="openapi_context">
+                    <attribute name="summary">Re-send the gift card email.</attribute>
```
loevgaard:
```suggestion
                    <attribute name="summary">Resend the gift card email to the customer.</attribute>
```

loevgaard (approved):
Just a few comments :)

src/Api/Controller/Action/ResendGiftCardEmailAction.php
Roshyo: No, it has to be named `$data` according to API Platform: https://api-platform.com/docs/core/controllers/

> Warning: the __invoke() method parameter MUST be called $data, otherwise, it will not be filled correctly!

src/Api/Controller/Action/ResendGiftCardEmailAction.php
Roshyo: Yeah true. I C/P from existing controller, but I will refactor to that indeed

src/Api/Controller/Action/ResendGiftCardEmailAction.php
loevgaard: Okay, cool. But don't copy it to a new variable then, imo. For such small files it doesn't matter with naming, and we have type hints, static analysis and everything - we won't mess it up :D

PR #163, opened 2021-10-06
Add routes permissions for admin

Roshyo:


loevgaard (approved):

igormukhingmailcom:
Few notes:
- That parameter is used only by `ResourceController` so our custom actions just ignoring it
- Even using it with the original `ResourceController` does not lead to any effect as the only service that implements `AuthorizationCheckerInterface` - always returns `true`. Looks like that was designed for Sylius Plus (Rbac).
- Admin permissions are checking by default for any route if that route's path starts with `/admin/` (or what we have at config)

Roshyo:
Yes, that's exactly my need :D

I'm using it on an app that has some kind of right management, and it was ignored because of that

PR #173, opened 2021-11-16
Make 0.11 branch compatible with Sylius 1.9

### LINKED ISSUE #160: Can't seem to install any version of this plugin in sylius 1.8 or 1.9
We are in the process of upgrading our sylius store from 1.8 to 1.9, our 1.8 branch is running on symfony 4.4, our 1.9 branch is running symfony 5.2.

Unfortunately I have not been able to install this plugin in either branch because the dependencies can't be resolved.

I started looking through the history of composer.json for th plugin and it seems that the dependencies have changes quite drastically between tagged releases, and it's not obvious which versions of sylius the plugin targets... some of the deps seem too new for S1.8 but then if i try to install it on S1.9, requires a lower version (^2.8) of the REST Bundle(v3.0.5) than our store (Sylius1.9.6)

Roshyo:
Fixes #160

Merge this OR #172  but not both of them.


loevgaard:
I think this PR can be closed?

PR #185, opened 2022-01-17
Allow to define custom rendering options for wkhtmltopdf

### LINKED ISSUE #184: Allow to define PDF rendering options
It would be nice to be able to define some PDF options for rendering, such as:

* Format (A4, A5, ...)
* Mode (Portrait/Landscape)

Roshyo:
Fixes #184


composer.json
```diff
@@ -70,6 +70,12 @@
         ]
     },
     "config": {
+        "allow-plugins": {
+            "composer/package-versions-deprecated": true,
+            "dealerdirect/phpcodesniffer-composer-installer": true,
```
loevgaard: Remove this

src/DependencyInjection/Configuration.php
```diff
@@ -37,6 +37,55 @@ public function getConfigTreeBuilder(): TreeBuilder
         $rootNode
             ->addDefaultsIfNotSet()
             ->children()
+                ->arrayNode('pdf_rendering')
+                ->addDefaultsIfNotSet()
+                    ->children()
+                        ->scalarNode('default_orientation')->defaultValue('Portrait')->end()
+                        ->arrayNode('available_orientations')
```
loevgaard, line 50: No need for this I think

src/DependencyInjection/Configuration.php
```diff
@@ -37,6 +37,55 @@ public function getConfigTreeBuilder(): TreeBuilder
         $rootNode
             ->addDefaultsIfNotSet()
             ->children()
+                ->arrayNode('pdf_rendering')
+                ->addDefaultsIfNotSet()
+                    ->children()
+                        ->scalarNode('default_orientation')->defaultValue('Portrait')->end()
```
loevgaard:
```suggestion
                        ->scalarNode('orientation')->defaultValue('Portrait')->end()
```

Could we replace `'Portrait'` with a constant?

Can you add info? I.e. `The default orientation for rendered gift cards`

src/DependencyInjection/Configuration.php
```diff
@@ -37,6 +37,55 @@ public function getConfigTreeBuilder(): TreeBuilder
...
+                        ->arrayNode('available_orientations')
+                            ->scalarPrototype()->end()
+                            ->defaultValue([
+                                'Portrait',
+                                'Landscape',
+                            ])
+                        ->end()
+                        ->scalarNode('default_page_size')->defaultValue('A4')->end()
```
loevgaard:
```suggestion
                        ->scalarNode('page_size')->defaultValue('A4')->end()
```

Same comment as above

src/Generator/GiftCardPdfGenerator.php
```diff
@@ -16,10 +17,16 @@ class GiftCardPdfGenerator implements GiftCardPdfGeneratorInterface
...
     private GeneratorInterface $snappy;
 
-    public function __construct(Environment $twig, GeneratorInterface $snappy)
-    {
+    private PdfRenderingOptionProviderInterface $renderingOptionProvider;
+
+    public function __construct(
+        Environment $twig,
+        GeneratorInterface $snappy,
+        PdfRenderingOptionProviderInterface $renderingOptionProvider
```
loevgaard:
```suggestion
        PdfRenderingOptionsProviderInterface $renderingOptionsProvider
```

tests/Unit/DependencyInjection/ConfigurationTest.php
```diff
@@ -88,4 +128,23 @@ public function processed_configuration_for_array_node_1(): void
             ],
         ]);
     }
+
+    /**
+     * @test
+     */
+    public function it_overrides_default_configuration(): void
```
loevgaard: This belongs in an extension test

loevgaard:
Very nice PR, Stephane 🎉

psalm.xml
```diff
@@ -13,6 +13,7 @@
             <directory name="src/Fixture"/>
             <directory name="tests/Application"/>
             <directory name="vendor"/>
+            <file name="src/DependencyInjection/Configuration.php"/>
```
loevgaard: You should not ignore this. Just add the applicable suppressions inside the file

src/DependencyInjection/Configuration.php
Roshyo: We need at some point to allow extension of available parameters imo

src/DependencyInjection/Configuration.php
Roshyo: Where would you put that constant ?

loevgaard: Some class - not sure of the name of that class though :D

src/DependencyInjection/Configuration.php
loevgaard: Okay, but it won't be a breaking change to add this by then. Right now I can't think of other orientations than portrait and landscape

tests/Unit/DependencyInjection/ConfigurationTest.php
Roshyo: Not sure about how to proceed

src/DependencyInjection/Configuration.php
Roshyo: I would prefer to keep it, for consistency reasons (we have that for format as well), and also because it does not add complexity at all, and may allow any dev to simply disable one orientation (w/e their reasons). What do you think ?

psalm.xml
Roshyo: It raises a lot of new errors that were not present before. I suspect some update from Psalm or Symfony.

Errors like this: 

```
ERROR: PossiblyUndefinedMethod - src/DependencyInjection/Configuration.php:47:27 - Method Symfony\Component\Config\Definition\Builder\NodeDefinition::arrayNode does not exist (see https://psalm.dev/108)
                        ->arrayNode('available_orientations')
```

And there are another 130 errors like this one... So I'm not sure about how to proceed

loevgaard: Like this: https://github.com/Setono/SyliusImagePlugin/blob/master/src/DependencyInjection/Configuration.php#L27

src/DependencyInjection/Configuration.php
loevgaard: Okay 👍

src/DependencyInjection/SetonoSyliusGiftCardExtension.php
```diff
@@ -15,13 +15,39 @@ final class SetonoSyliusGiftCardExtension extends AbstractResourceExtension
     public function load(array $configs, ContainerBuilder $container): void
     {
         /**
-         * @var array{code_length: int, driver: string, resources: array<string, mixed>} $config
+         * @var array{
+         *     pdf_rendering: array{
+         *         default_orientation: string,
+         *         available_orientations: array<array-key, string>,
```
loevgaard:
```suggestion
         *         available_orientations: list<string>,
```
I have started to use `list` where applicable just to get used to it. It adds another layer of static analysis and hence safety :)

src/Validator/Constraints/Pdf/ValidPageSizeValidator.php
```diff
@@ -0,0 +1,35 @@
...
+        $this->availablePageSizes = $availableOrientations;
+    }
+
+    public function validate($value, Constraint $constraint): void
+    {
+        if (!$constraint instanceof ValidPageSize) {
+            throw new UnexpectedTypeException($constraint, ValidPageSize::class);
+        }
+
+        if (null === $value) {
```
loevgaard: Shouldn't validators return for both `null` _and_ `''` because the `NotBlank` validation will take care of those cases?

tests/Unit/DependencyInjection/ConfigurationTest.php
loevgaard: https://github.com/SymfonyTest/SymfonyDependencyInjectionTest#testing-a-container-extension

src/Form/Type/GiftCardConfigurationType.php
```diff
@@ -7,12 +7,32 @@
...
 use Symfony\Component\Form\CallbackTransformer;
 use Symfony\Component\Form\Extension\Core\Type\CheckboxType;
+use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
 use Symfony\Component\Form\Extension\Core\Type\CollectionType;
 use Symfony\Component\Form\Extension\Core\Type\TextType;
 use Symfony\Component\Form\FormBuilderInterface;
 
 final class GiftCardConfigurationType extends AbstractResourceType
 {
+    private array $availableOrientations;
```
loevgaard, line 18: I would usually phpdoc arrays, like `list<string>`

Issue #187, opened 2022-01-17
Preview PDF rendering in admin + allow CSS edit


Roshyo:
For better UX, we should:

* Allow user to update CSS from the admin panel
* Add a default CSS so user can fallback to it if needed

PR #191, opened 2022-01-22
Allow live pdf template edit

Roshyo:
Related to #187 

image


composer.json
```diff
@@ -70,7 +70,13 @@
         ]
     },
     "config": {
-        "sort-packages": true
+        "sort-packages": true,
+        "allow-plugins": {
+            "composer/package-versions-deprecated": true,
+            "dealerdirect/phpcodesniffer-composer-installer": true,
```
loevgaard: We don't need this to run I think

src/Controller/Action/AdminGeneratePdfAction.php
```diff
@@ -0,0 +1,51 @@
...
+use Setono\SyliusGiftCardPlugin\Factory\GiftCardFactoryInterface;
+use Setono\SyliusGiftCardPlugin\Form\Type\GiftCardConfigurationType;
+use Setono\SyliusGiftCardPlugin\Generator\GiftCardPdfGeneratorInterface;
+use Setono\SyliusGiftCardPlugin\Model\GiftCardConfigurationInterface;
+use Sylius\Component\Resource\Repository\RepositoryInterface;
+use Symfony\Component\Form\FormFactoryInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+
+final class AdminGeneratePdfAction
```
loevgaard: Put inside `Admin` folder instead?

src/Factory/GiftCardFactoryInterface.php
```diff
@@ -24,4 +24,6 @@ public function createFromOrderItemUnitAndCart(
         OrderItemUnitInterface $orderItemUnit,
         OrderInterface $cart
     ): GiftCardInterface;
+
+    public function createDummy(): GiftCardInterface;
```
loevgaard: What is this used for? :) Could we have a better name?

src/Generator/GiftCardPdfGenerator.php
```diff
@@ -33,4 +39,21 @@ public function generatePdfResponse(
 
         return new PdfResponse($this->snappy->getOutputFromHtml($html), 'gift_card.pdf');
     }
+
+    public function generateAndSavePdf(
```
loevgaard, line 68: Why isn't this method returning anything? :)

Roshyo: It is just generating the PDF and creating the file. So nothing to return here imo.

loevgaard: But wouldn't you expect from the name of the method that you either give the filename as input or it returns the filename as output?

Roshyo: Well, Since the file path generation is handled in another service, I'm not sure it is necessary. But we can return the name indeed.

src/Controller/Action/Admin/RenderPdfAction.php
```diff
@@ -0,0 +1,53 @@
...
+        RepositoryInterface $giftCardConfigurationRepository,
+        GiftCardPdfPathGeneratorInterface $giftCardPdfPathGenerator,
+        Filesystem $filesystem
+    ) {
+        $this->giftCardConfigurationRepository = $giftCardConfigurationRepository;
+        $this->giftCardPdfPathGenerator = $giftCardPdfPathGenerator;
+        $this->filesystem = $filesystem;
+    }
+
+    public function __invoke(Request $request, int $id): PdfResponse
```
loevgaard: I would have expected that the `$id` was the id of the gift card, but it's not apparently. Can you explain the flow of this. I don't seem to get it

src/Factory/DummyGiftCardFactory.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Setono\SyliusGiftCardPlugin\Factory;
+
+use Setono\SyliusGiftCardPlugin\Model\GiftCardInterface;
+use Sylius\Component\Currency\Context\CurrencyContextInterface;
+
+final class DummyGiftCardFactory implements DummyGiftCardFactoryInterface
```
loevgaard: Can you explain the usage of this 'DummyGiftCard' ?

src/Resources/config/doctrine/model/GiftCardConfiguration.orm.xml
```diff
@@ -14,6 +14,7 @@
             </options>
         </field>
         <field name="defaultValidityPeriod" nullable="true" />
+        <field name="pdfRenderingCss" type="text" nullable="true" />
```
loevgaard, line 19: What is this? Can I input CSS that will be applied to the gift card?

src/Controller/Action/Admin/RenderPdfAction.php
Roshyo: Well, we are in the admin, configuring a GiftCardConfiguration. We don't have any gift card right now (hence the dummy factory) but we just want to render the PDF that will be given to any user buying a gift card with this configuration

src/Factory/DummyGiftCardFactory.php
Roshyo: See above

src/Resources/config/doctrine/model/GiftCardConfiguration.orm.xml
Roshyo: Yes, exactly. No linter on this right now though :D

src/Resources/views/Admin/GiftCardConfiguration/Update/_javascripts.html.twig
```diff
@@ -0,0 +1,6 @@
+<script src="{{ asset('bundles/setonosyliusgiftcardplugin/setono-sylius-gift-card-live-pdf-rendering.js') }}"></script>
```
loevgaard, line 1: What's the advantage of having a file instead of just having the JS right here?

Roshyo: We should never put JS directly into HTML imo. It is not the best practice

src/Resources/views/Admin/GiftCardConfiguration/_form.html.twig
```diff
@@ -23,3 +23,33 @@
...
+
+            <button class="ui button primary labeled icon js-ssgc-apply-pdf-changes"
+                    data-url="{{ path('setono_sylius_gift_card_admin_ajax_generate_pdf', {'id': resource.id}) }}">
+                <i class="icon check"></i>
+                {{ 'setono_sylius_gift_card.ui.apply'|trans }}
+            </button>
+        </div>
+        <div class="column js-ssgc-live-render-container" style="height: 900px;">
+            <embed id="js-ssgc-live-render-box"
+                   src="{{ url('setono_sylius_gift_card_admin_partial_render_pdf', {id: resource.id}) }}"
```
loevgaard: Imo you should just fetch the PDF on page load so that you don't need two endpoints

src/Controller/Action/Admin/RenderPdfAction.php
loevgaard: Of course, haha. Well, the name of this action could be better, i.e. `RenderExamplePdfAction`

src/Factory/DummyGiftCardFactory.php
loevgaard: I think `Example` is a better name than `Dummy`. Dummy in my vocabulary really means something you shouldn't use in production

src/Resources/views/Admin/GiftCardConfiguration/_form.html.twig
Roshyo: There were issues with the embed thing and dynamically generating the PDF, hence this flow

src/Factory/DummyGiftCardFactory.php
Roshyo: Here is the definition I use: `an object designed to resemble and serve as a substitute for the real or usual one.`

Which is clearly what it does :thinking: `Example` in my mind would read as the same as `CustomerExampleFactory` in fixtures, and i'm not sure it is what we intend

src/Resources/views/Shop/GiftCard/pdf.html.twig
```diff
@@ -9,16 +9,7 @@
...
-                font-family: sans-serif;
-                font-size: 3rem;
-                background: transparent url("{{ path }}") no-repeat left bottom;
-                background-size: 100px;
-            }
-
-            img {
-                display: block;
-            }
+            {{ configuration.pdfRenderingCss|default('')|ssgc_replace_css_option(_context)|raw }}
```
loevgaard, line 12: How do we use some kind of default value if the user hasn't added any CSS?

We are also still missing the output of the gift card HTML on the gift card configuration page so that users easily can see the selectors they can use in the CSS, right?

src/Factory/DummyGiftCardFactory.php
loevgaard: Exactly, a dummy is not something that should be used for real :D But this example is _real_ :D

src/Resources/views/Shop/GiftCard/pdf.html.twig
Roshyo: Right, there is no default value as of now.
Neither display of the HTML.
I'll add that.

composer-require-checker.json
```diff
@@ -77,6 +77,7 @@
     "Sylius\\Component\\Product\\Resolver\\ProductVariantResolverInterface",
     "Sylius\\Component\\Promotion\\Checker\\Rule\\RuleCheckerInterface",
     "Sylius\\Component\\Promotion\\Model\\PromotionSubjectInterface",
-    "Sylius\\Component\\User\\Model\\UserInterface"
+    "Sylius\\Component\\User\\Model\\UserInterface",
+    "Symfony\\Component\\DomCrawler\\Crawler"
```
loevgaard: Shouldn't this be added to composer.json?

src/Factory/ExampleGiftCardFactory.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Setono\SyliusGiftCardPlugin\Factory;
+
+use Setono\SyliusGiftCardPlugin\Model\GiftCardInterface;
+use Sylius\Component\Currency\Context\CurrencyContextInterface;
+
+final class ExampleGiftCardFactory implements ExampleGiftCardFactoryInterface
```
loevgaard: Why don't we have a method called `createExample()` on the `GiftCardFactoryInterface`?

src/Generator/GiftCardPdfGenerator.php
```diff
@@ -16,10 +17,20 @@ class GiftCardPdfGenerator implements GiftCardPdfGeneratorInterface
...
-    {
+    private GiftCardPdfPathGeneratorInterface $giftCardPdfPathGenerator;
+
+    private Filesystem $filesystem;
+
+    public function __construct(
+        Environment $twig,
+        GeneratorInterface $snappy,
+        GiftCardPdfPathGeneratorInterface $giftCardPdfPathGenerator,
+        Filesystem $filesystem
```
loevgaard: Use interface?

src/Resources/config/routes/admin_partial.yaml
```diff
@@ -0,0 +1,6 @@
+setono_sylius_gift_card_admin_partial_render_pdf:
+    path: /gift-card-configurations/{id}/render-pdf
+    controller: setono_sylius_gift_card.controller.action.admin_render
+    methods: [GET, POST, PUT]
```
loevgaard: GET, POST, PUT? Do we need all of those?

src/Resources/config/routes/admin_ajax.yaml
```diff
@@ -24,3 +24,17 @@ setono_sylius_gift_card_admin_ajax_customer_by_email:
             repository:
                 method: findBy
                 arguments: [email: $email]
+
+setono_sylius_gift_card_admin_ajax_generate_pdf:
+    path: /gift-card-configurations/{id}/generate-pdf
+    methods: [ GET, POST, PUT ]
```
loevgaard: GET, POST, PUT? Do we need all of those?

src/Resources/private/default-gift-card-configuration-css.txt
```diff
@@ -0,0 +1,10 @@
+body {
```
loevgaard: Why not name this file `.css`?

src/Factory/ExampleGiftCardFactory.php
Roshyo: Because that way, it is clear that you have an ExampleGiftCard.
I think it is cleaner than having just a method in regular factory.

src/Resources/private/default-gift-card-configuration-css.txt
Roshyo: To make it clear that it is no CSS loaded the regular way

src/Controller/Action/Admin/GenerateExamplePdfAction.php
```diff
@@ -0,0 +1,53 @@
...
+use Setono\SyliusGiftCardPlugin\Form\Type\GiftCardConfigurationType;
+use Setono\SyliusGiftCardPlugin\Generator\GiftCardPdfGeneratorInterface;
+use Setono\SyliusGiftCardPlugin\Model\GiftCardConfigurationInterface;
+use Sylius\Component\Resource\Repository\RepositoryInterface;
+use Symfony\Component\Form\FormFactoryInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Webmozart\Assert\Assert;
+
+final class GenerateExamplePdfAction
```
loevgaard: So either this action should return the PDF and we will encode it client side (with https://developer.mozilla.org/en-US/docs/Web/API/btoa) or this action should be renamed to `GenerateEncodedExamplePdfAction` I think. Wdyt?

src/Resources/private/default-gift-card-configuration-css.txt
loevgaard: Should this file be a `.css.twig` file and be rendered using twig? Wdyt?

src/Factory/GiftCardFactoryInterface.php
```diff
@@ -24,4 +24,6 @@ public function createFromOrderItemUnitAndCart(
         OrderItemUnitInterface $orderItemUnit,
         OrderInterface $cart
     ): GiftCardInterface;
+
```
loevgaard, line 27: Maybe add a comment here about what it's used for ;)

src/Resources/private/default-gift-card-configuration-css.txt
Roshyo: Absolutely. And it simplifies some parts also :)

loevgaard (approved):

Issue #193, opened 2022-02-03
Purchased card has a zero amount


jum4:
On 0.12.x branch, when a gift card is automatically created during a purchase, the amount is filled only if the card amount is configurable.
For fixed amount, the amount is allways 0.

I think it is because the card is created in the `AddToCartTypeExtension`, and the cartItem prices are not already calculated at this step.

I can suggest a working pull request, but I am not sure about the method to use

Roshyo:
I don't really get the problem here, as if the gift card has not a configurable amount (so fixed one) then Sylius mechanics takes place and set the OrderItem price itself.

What is your issue exactly ?

jum4:
My issue is the gift card is actually created with a 0 amount, maybe I misconfigured something ?

Roshyo:
The gift card amount should be done on OrderProcess. Can you come up with either some test to reproduce, or a scenario ?

jum4:
GiftCardPlugin: 0.12.x
Sylius: 1.10
* create a product with "gift-card" ON and "amount configurable" OFF
* add the product to the cart
* a gift cart is created with amount 0€

Roshyo:
Thank you @jum4 , I could reproduce with your scenario, and your PR is indeed fixing the issue. I'm just adding tests and we'll be able to proceed with merge

PR #194, opened 2022-02-03
Set cartItem unit price before giftcard creation

jum4:
Related to #193

### DIFF
--- a/src/Form/Extension/AddToCartTypeExtension.php
+++ b/src/Form/Extension/AddToCartTypeExtension.php
@@ -76,8 +76,9 @@ final class AddToCartTypeExtension extends AbstractTypeExtension
             return;
         }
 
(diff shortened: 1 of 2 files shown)


Roshyo (approved):

loevgaard:
@Roshyo: How did this work in `0.11`? Why has it changed? Do you know that?

Roshyo:
It stopped working when we started to allow GC with custom amounts.
Previously, we were creating the GC from the OrderItemUnit when placing the order. Now we do it when creating the Order (cart) and the Order Item Unit has no price yet, since it is assigned by Sylius later.
Perhaps an other fix would be to change the GC creation just after this price assignment (from order processor) but not sure if it would raise more problems or not.

loevgaard:
Thank you @jum4 and @Roshyo 🎉

PR #199, opened 2022-03-03
Make actions compatible with newer Sylius versions

### LINKED ISSUE #197: 0.11.7 Incompatible with friendsofsymfony/rest-bundle ^3.0
Using the 0.11.7 version of SyliusGiftCardPlugin.

It requires friendsofsymfony/rest-bundle: "^2.8 || ^3.0".

However FOS\RestBundle\View\View does not have a setTemplate method anymore since 3.0
When I open my cart summary in the frontend, I get this:
**Attempted to call an undefined method named "setTemplate" of class "FOS\RestBundle\View\View".
in vendor/setono/sylius-gift-card-plugin/src/Controller/Action/AddGiftCardToOrderAction.php (line 80)**

Requiring "^2.8" does not work as 2.8 does not support PHP 8.

I don't have much experience with the REST bundle. Why is it needed?
Couldn't the Action simply extend symfony's AbstractController and use $this->render($template, $context); like every normal Action?

loevgaard:
Fixes https://github.com/Setono/SyliusGiftCardPlugin/issues/197


composer.json
```diff
@@ -32,6 +31,7 @@
     },
     "require-dev": {
         "doctrine/doctrine-bundle": "^1.2.13",
+        "friendsofsymfony/rest-bundle": "^2.8",
```
loevgaard: Maybe we should add `^2.8 || ^3.0` and then add it to `require` instead?

src/Controller/Action/SearchGiftCardAction.php
```diff
@@ -10,35 +10,47 @@
...
 
     private FormFactoryInterface $formFactory;
 
-    public function __construct(
-        ViewHandlerInterface $viewHandler,
-        FormFactoryInterface $formFactory
-    ) {
+    private ?Environment $twig;
+
+    public function __construct(?ViewHandlerInterface $viewHandler, FormFactoryInterface $formFactory, Environment $twig = null)
```
Roshyo: put the arguments on 3 lines

Roshyo (changes requested):

Roshyo (approved):

PR #219, opened 2022-05-06
Allow preventing sending the notification email on GC creation

### LINKED ISSUE #205: Allow administrator to choose whether to send a notification or not
This is the current create/update form for gift cards. The information message reads:

> When you associate a customer with a gift card, Sylius will automatically notify this customer via email

image

It would make much more sense if we added a checkbox that said: `[ ] Notify customer` (and only show that checkbox when a customer is chosen).

Roshyo:
Fixes #205

image

image

### DIFF
--- a/src/Model/GiftCard.php
+++ b/src/Model/GiftCard.php
@@ -49,6 +49,8 @@ class GiftCard implements GiftCardInterface
 
     protected ?DateTimeInterface $expiresAt = null;
 
+    protected bool $sendNotificationEmail = true;
+
     public function __construct()
     {
         $this->appliedOrders = new ArrayCollection();
@@ -295,5 +297,15 @@ class GiftCard implements GiftCardInterface
         }
 
         return $date > $giftCardValidUntil;
+    }
+
+    public function getSendNotificationEmail(): bool
+    {
+        return $this->sendNotificationEmail;
+    }
+
+    public function setSendNotificationEmail(bool $sendNotificationEmail = true): void
+    {
+        $this->sendNotificationEmail = $sendNotificationEmail;
     }
 }
--- a/src/Resources/public/setono-sylius-gift-card-send-notification-email.js
+++ b/src/Resources/public/setono-sylius-gift-card-send-notification-email.js
@@ -1,0 +1,32 @@
+(function ($) {
+  'use strict';
+
+  $.fn.extend({
+    sendNotificationEmailInput: function () {
+      const $customerInput = $(this);
+      const $sendCustomerNotificationEmail = $($customerInput.data('target'));
+
+      $('label[for="setono_sylius_gift_card_gift_card_customer"]').prepend('<i class="delete icon js-ssgc-clear-input" style="float: right;cursor: pointer"></i>');
+
+      $customerInput.on('change', () => {
+        if ($customerInput.val() !== '') {
+          $sendCustomerNotificationEmail.css({
+            'visibility': 'visible',
+            'height': 'auto',
+          });
+        } else {
+          $sendCustomerNotificationEmail.css({
+            'visibility': 'hidden',
+            'height': '0',
+          });
+        }
+      });
+
+      $('.js-ssgc-clear-input').on('click', () => {
+        $customerInput.parent().children('div.text').html('');
+        $customerInput.val('');
+        $customerInput.trigger('change');
+      });
+    }
+  });
+})(jQuery);
--- a/src/Resources/public/setono-sylius-gift-card-send-notification-email.js
+++ b/src/Resources/public/setono-sylius-gift-card-send-notification-email.js
@@ -6,6 +6,12 @@
       const $customerInput = $(this);
       const $sendCustomerNotificationEmail = $($customerInput.data('target'));
 
+      if ($customerInput.val() !== '') {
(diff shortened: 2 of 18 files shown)


src/Resources/translations/messages.da.yml
```diff
@@ -20,6 +20,7 @@ setono_sylius_gift_card:
             channel: Kanal
             custom_message: Personlig besked
             expires_at: Udløber
+            send_notification_email: Send notification email
```
loevgaard:
```suggestion
            send_notification_email: Send email med gavekort til kunde
```

src/Resources/views/Admin/GiftCard/Create/_javascripts.html.twig
```diff
@@ -0,0 +1,6 @@
+<script src="{{ asset('bundles/setonosyliusgiftcardplugin/setono-sylius-gift-card-send-notification-email.js') }}"></script>
+<script>
+    $().ready(() => {
```
loevgaard:
```suggestion
    $(() => {
```

src/Resources/public/setono-sylius-gift-card-send-notification-email.js
```diff
@@ -0,0 +1,32 @@
...
+  $.fn.extend({
+    sendNotificationEmailInput: function () {
+      const $customerInput = $(this);
+      const $sendCustomerNotificationEmail = $($customerInput.data('target'));
+
+      $('label[for="setono_sylius_gift_card_gift_card_customer"]').prepend('<i class="delete icon js-ssgc-clear-input" style="float: right;cursor: pointer"></i>');
+
+      $customerInput.on('change', () => {
+        if ($customerInput.val() !== '') {
+          $sendCustomerNotificationEmail.css({
```
loevgaard: What's wrong with just `show()` and `hide()`? :D

src/Resources/views/Admin/GiftCard/_form.html.twig
```diff
@@ -6,8 +6,13 @@
             {{ form_row(form.code) }}
             {{ form_row(form.enabled) }}
             {{ form_row(form.expiresAt) }}
-            <div class="ui info message">{{ 'setono_sylius_gift_card.ui.customer_help_text'|trans }}</div>
-            {{ form_row(form.customer) }}
+            {{ form_row(form.customer, {'attr': {'data-target': '.js-ssgc-send-customer-email'}}) }}
+            {% if form.sendNotificationEmail is defined %}
+                <div class="js-ssgc-send-customer-email" style="height: 0; visibility: hidden;">
```
loevgaard:
```suggestion
                <div class="js-ssgc-send-customer-email" style="display:none">
```
?

src/Resources/views/Admin/GiftCard/Create/_javascripts.html.twig
```diff
@@ -0,0 +1,6 @@
+<script src="{{ asset('bundles/setonosyliusgiftcardplugin/setono-sylius-gift-card-send-notification-email.js') }}"></script>
+<script>
+    $(() => {
```
loevgaard, line 3: Can you check if this worked?

I know there is a shorthand. You can read about it here: https://learn.jquery.com/using-jquery-core/document-ready/

src/Resources/views/Admin/GiftCard/_form.html.twig
Roshyo: Display none will make it disappear from the DOM and it may lead to undesired side effects as the form value would not be submitted at all. So I prefer this approach

Roshyo: Display none will make it disappear from the DOM and it may lead to undesired side effects as the form value would not be submitted at all. So I prefer this approach

src/Resources/views/Admin/GiftCard/Create/_javascripts.html.twig
Roshyo: Yes it does

src/Resources/views/Admin/GiftCard/_form.html.twig
loevgaard: Yes, but we also have a default value on the entity right, so this is actually what we want, right?

loevgaard:
I think what this PR is missing is the case where an administrator updates a gift card, but doesn't change the email, then the user would get an email notification again.

This could easily be solved, I think, by just having the `sendNotificationEmail` property to be `false` by default

src/Model/GiftCard.php
```diff
@@ -49,6 +49,8 @@ class GiftCard implements GiftCardInterface
 
     protected ?DateTimeInterface $expiresAt = null;
 
+    protected bool $sendNotificationEmail = true;
```
loevgaard, line 52: I think this should be `false` by default. See my comment here: https://github.com/Setono/SyliusGiftCardPlugin/pull/219#issuecomment-1119427019

Roshyo: No, he would not since the email is on gc_create event

loevgaard: Okay. Great. Thanks

loevgaard (approved):

PR #221, opened 2022-05-06
Make GiftCardConfigurationProvider always return GiftCardConfiguration

### LINKED ISSUE #220: GiftCardConfigurationProvider refactor
In my opinion the `GiftCardConfigurationProvider` which is called `GiftCardChannelConfigurationProvider` in the plugin in `0.12.x` as of right now should always return a gift card configuration, no matter what. If no configuration exists for the given channel and locale it should create one and persist it.

loevgaard:
Fixes #220 

I think this PR makes a lot of sense given the name of the provider. You expect the `GiftCardConfigurationProvider` to actually provide a `GiftCardConfiguration`.


Roshyo (approved):
I like this idea, it will ease the work to configure the plugin

PR #229, opened 2022-05-13
Add a command to delete old gift cards

### LINKED ISSUE #228: Create a command to delete old gift cards
It would be handy to have a command that would delete expired gift cards automatically (where `expiresAt` is in the the past)

Also, since we would need that feature in 0.11, it would make sense to base this on `createdAt` property for that branch

Roshyo:
Fixes #228

### DIFF
--- a/src/Command/Cli/DeleteOldGiftCardsCommand.php
+++ b/src/Command/Cli/DeleteOldGiftCardsCommand.php
@@ -33,6 +33,16 @@ final class DeleteOldGiftCardsCommand extends Command
     protected function configure(): void
     {
         $this->setDescription('Deletes gift cards older than the provided date.');
+        $this->setHelp(<<<'EOF'
+The <info>%command.name%</info> command deletes gift cards older than the provided date.
+
+Date option accepts any date format accepted by PHP's <info>DateTime</info> class.
+Such as "2020-01-01" or "2020-01-01 12:00:00" or "-3 years".
+
+Period option accepts any period format accepted by PHP's <info>DateInterval</info> class.
+Such as "P1Y" or "P1Y2M".
+EOF
+        );
         $this->addOption(
             'date',
             null,
@@ -67,7 +77,12 @@ final class DeleteOldGiftCardsCommand extends Command
         }
 
         try {
-            $thresholdDate = new \DateTimeImmutable($date ?? $period);
+            if (null !== $date) {
+                $thresholdDate = new \DateTimeImmutable($date);
+            } else {
+                $thresholdDate = new \DateTimeImmutable('now');
+                $thresholdDate = $thresholdDate->sub(new \DateInterval($period));
+            }
         } catch (\Exception $e) {
             $output->writeln('The provided date or period is not valid.');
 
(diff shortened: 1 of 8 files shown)


src/Doctrine/ORM/GiftCardRepository.php
```diff
@@ -69,4 +69,14 @@ public function createAccountListQueryBuilder(CustomerInterface $customer): Quer
 
         return $qb;
     }
+
+    public function findCreatedBefore(\DateTimeInterface $date): array
+    {
+        $qb = $this->createQueryBuilder('gc');
+
+        $qb->andWhere('gc.createdAt < :date');
+        $qb->setParameter('date', $date->format('Y-m-d H:i:s'));
```
loevgaard:
```suggestion
        $qb->setParameter('date', $date);
```

Isn't this possible?

src/Command/Cli/DeleteOldGiftCardsCommand.php
```diff
@@ -0,0 +1,103 @@
...
+    protected static $defaultName = 'setono:sylius-gift-card:delete-old-gift-cards';
+
+    private GiftCardRepositoryInterface $giftCardRepository;
+
+    private ObjectManager $giftCardManager;
+
+    public function __construct(
+        GiftCardRepositoryInterface $giftCardRepository,
+        ObjectManager $giftCardManager,
+        ?string $name = null
```
loevgaard:
```suggestion
```

Why do you need the name?

Roshyo: To call parent constructor

src/Command/Cli/DeleteOldGiftCardsCommand.php
```diff
@@ -0,0 +1,103 @@
...
+            }
+        } catch (\Exception $e) {
+            $output->writeln('The provided date or period is not valid.');
+
+            return 0;
+        }
+
+        $giftCards = $this->giftCardRepository->findCreatedBefore($thresholdDate);
+        foreach ($giftCards as $giftCard) {
+            $this->giftCardManager->remove($giftCard);
```
loevgaard: I think this command should instead disable the gift cards and not remove them

src/Doctrine/ORM/GiftCardRepository.php
```diff
@@ -69,4 +69,14 @@ public function createAccountListQueryBuilder(CustomerInterface $customer): Quer
 
         return $qb;
     }
+
+    public function findCreatedBefore(\DateTimeInterface $date): array
```
loevgaard:
```suggestion
    public function findCreatedBefore(\DateTimeInterface $date, int $limit = 100): array
```
Else we could run into memory issues

Roshyo: I'd advise to allow null so we can get them all if we want to

src/Command/Cli/DeleteOldGiftCardsCommand.php
loevgaard: Yeah, but it's null by default, so you don't need it in your constructor :)

loevgaard (approved):

PR #234, opened 2022-06-02
Allow to preview uploaded image on GC config

### LINKED ISSUE #225: When I select an image it doesn't show a preview or anything else
Look at this video:

https://user-images.githubusercontent.com/2412177/167361209-3507451a-d0c7-44bd-ab15-6b05162e86cb.mov

When I select an image it doesn't show a preview of the image or anything else. This is a bad user experience.

Roshyo:
Fixes #225


src/Resources/views/Admin/GiftCardConfiguration/_javascripts.html.twig
```diff
@@ -0,0 +1,5 @@
+<script>
+    $().ready(() => {
```
loevgaard:
```suggestion
    $(() => {
```

loevgaard (approved):

PR #246, opened 2023-03-17
fix install doc

Florian-Merle:


README.md
```diff
@@ -169,7 +169,7 @@ use Sylius\Component\Core\Model\OrderItem as BaseOrderItem;
  * @ORM\Entity
  * @ORM\Table(name="sylius_order_item_unit")
```
loevgaard: I suppose this should be changed also?

Florian-Merle: I missed it :sweat_smile:

PR #257, opened 2023-11-08
Support sylius mailer 2

lruozzi9:


loevgaard:
Thanks, @lruozzi9. I will just improve the build a little and maybe make some deps more robust, and then I will release a new version :)

Issue #236, opened 2022-07-27
Disable buying of gift cards with gift cards


markbeazley:
At the moment it seems that it is kind of possible to order a gift card with another gift card, the customer can complete the checkout process and the old card is shown as used, but as the order has no payment it seems it doesn't enable the new card. This could allow for a user to refresh a gift card that is due to expire, although it would require an admin/staff user to manually enable to new card, its still not the best user experience. I can think of two possible solutions:

1. Catch this at the point the user tires to add the gift cart, and reject applying the card if their cart contains a gift card, then anytime the cart is updated remove any applied gift cards from teh order if the cart now contains a gift card.
2. Don't include the cost of gift cards when calculation the cost to use from the gift card, this would allow the customer to buy a new gift card and buy non gift card products in the same order whilst allowing them to use another gift card to pay for the non gift card products.

I think 2 might be the better solution, especially if it included a message/alert explaining that gift cards can't be used against gift cards.

loevgaard:
Oftentimes the simplest solution proves to be the best solution. Could you make a PR for option 1? :)

Issue #94, opened 2020-08-19
Add dashes to gift card codes to make them easier to read


loevgaard:
Instead of having a code like `AKSADFUBMSUA2CSA`, at code could look like `AKSA-DFUB-MSUA-2CSA`. This could also be implemented when displaying the code and then removing anything else but letters and numbers when inputting the code.

PR #269, opened 2026-07-03
1.x rewrite (Sylius 1.13/1.14, no API layer)

loevgaard:
## Summary

This is the ground-up **1.x rewrite** of the plugin, targeting **Sylius 1.13/1.14**, **PHP ≥ 8.1**, **Symfony ^6.4**. It replaces the `0.12.x` architecture (base branch of this PR).

The rewrite simplifies aggressively and hardens what remains — see [`REWRITE.md`](REWRITE.md) for the full architecture, decisions and progress log, and [`UPGRADE-1.0.md`](UPGRADE-1.0.md) for the migration inventory.

## Highlights

- **No API layer** (API Platform, serializers, JWT, voters all removed).
- **One gift card type** — the customer chooses the amount (configurable min/max); **virtual vs physical is derived from `variant->isShippingRequired()`**, not a special product type.
- **Purchase flow**: a disabled "pending" gift card is created per `OrderItemUnit` at add-to-cart, reconciled to final amounts at checkout completion, enabled + emailed (PDF attachment) on payment, disabled on cancel.
- **Redemption is a payment, not a discount.** Each applied gift card becomes a completed `Payment` against the order via a lazily-created offline gift card payment method; the order total stays intact, the remainder goes through the normal gateway, and the payment step is skipped at full coverage. See [the note below](#why-redemption-is-a-payment).
- **Append-only `GiftCardTransaction` ledger** (audit + idempotency) which accounts for the *whole* balance, opening balance included — `sum(transactions) == amount`. Balance mutations go through a single balance operator; nothing below controllers flushes.
- **Both state machine adapters supported** — winzou callbacks *and* the equivalent Symfony Workflow listeners, so the plugin behaves the same whichever `sylius_core.state_machine.default_adapter` is set to.
- **Designs** (translatable, front/back images) with a live product-page preview that matches the dompdf-rendered PDF.
- **Admin**: gift card grid, design CRUD, one-click "create gift card product" scaffold, manual balance adjustments, outstanding-balance dashboard.
- **Tooling** modelled on Setono/SyliusPluginSkeleton 1.14.x: PHPStan (max), ECS, Rector, Infection, PHPUnit unit + functional suites, and a Playwright end-to-end suite. No Behat, no psalm.

## Why redemption is a payment

0.12.x reduced the order total with a negative `order_gift_card` adjustment. 1.x does not: a redeemed gift card is a payment against the order.

This started life as a configurable `redemption.mode` with both mechanisms, and was cut back to one after checking what other platforms actually do. None of them make it configurable, and the split is instructive: [Shopify](https://help.shopify.com/en/manual/products/gift-card-products/redeem-gift-card) and [BigCommerce](https://developer.bigcommerce.com/docs/rest-storefront/checkouts/checkout-gift-certificates) treat a gift card as a payment method, while [Magento](https://community.magento.com/t5/Magento-2-x-Feature-Requests-and/Gift-Certificates-should-apply-as-part-of-Total-Paid-on-orders/idi-p/465583) treats it as a discount and has a long-standing feature request to stop — because order management and accounting systems expect a payment instrument, and a gift card reducing the subtotal collides with promotions.

It also matches the accounting: selling a gift card takes money for a [deferred revenue liability](https://www.hubifi.com/blog/gift-cards-deferred-revenue), and redeeming it settles that liability rather than reducing what the order is worth.

**This is the sharpest upgrade edge in the PR**: orders no longer carry `order_gift_card` adjustments at all, so any host-app template, report or ERP export reading them silently returns nothing. `UPGRADE-1.0.md` calls it out.

## Testing

| | |
|---|---|
| PHPUnit | 51 unit + 12 functional |
| Playwright | 24 specs across 5 files, run in CI |
| CI | 8 jobs, PHP 8.1–8.3 × Sylius ~1.13/~1.14, lowest + highest deps |

The Playwright suite covers the admin (gift cards index/show/edit, designs, balance report, gift card and design preview PDFs, product edit for simple/configurable/gift card products) and the shop (gift card product page, locales, add-to-cart, and redemption in the cart). Specs discover their subjects through the admin grids and the locale switcher rather than hardcoding ids, so they survive a reseed.

## Notes for reviewers

- Large PR (52 commits) — organised roughly by rewrite phase (commit messages `1.x rewrite phase N …`) followed by polish, review follow-ups and bug fixes.
- The DI extension auto-configures state machines, grids, UI events, emails and image filters via `prepend()`; host apps only register the bundle, import routes and apply the entity traits.
- Deferred work is tracked as issues rather than left in `REWRITE.md`: #270, #271, #272, #273, #274, #275, #276.
- Some pre-existing open issues look resolved by this rewrite (e.g. #236, #94) but have deliberately been left alone until it merges.


.gitignore
```diff
@@ -2,12 +2,11 @@
...
-/behat.yml
-/phpspec.yml
+/.build/
 /.phpunit.result.cache
 .php-version
+
+/tests/Application/yarn.lock
+/.playwright-mcp/
+/tests/Application/public/build/
+/tests/Application/public/router.php
```
loevgaard: Use `symfony serve` instead of built-in php web server. Add this to CLAUDE.md

tests/Application/config/services_test.yaml
```diff
@@ -1,3 +1,3 @@
-imports:
```
loevgaard, line 1: This file should be removed

tests/Application/config/packages/webpack_encore.yaml
```diff
@@ -3,3 +3,4 @@ webpack_encore:
     builds:
         shop: '%kernel.project_dir%/public/build/shop'
         admin: '%kernel.project_dir%/public/build/admin'
+    strict_mode: false
```
loevgaard: This should not be added

src/Resources/translations/validators.en.yml
```diff
@@ -18,10 +18,9 @@ setono_sylius_gift_card:
...
-        background_image_required: The background image is required
-        default_configuration_must_exist: If no other configurations exist you must mark this configuration the default. Else you must mark another configuration as default before unchecking this
-    gift_card_search_command:
-        gift_card:
-            does_not_exist: The gift card with this code does not exist
-            not_blank: Please enter code
+    gift_card_design:
+        name:
+            not_blank: Please enter a name
+            max_length: The name must not be longer than {{ limit }} characters
```
loevgaard: The default validation messages are fine, so just remove these

src/Resources/config/prepend/winzou_state_machine.yaml
```diff
@@ -0,0 +1,36 @@
+winzou_state_machine:
```
loevgaard: Configuration that is prepended should be written as php arrays instead of loading files like this

.github/workflows/build.yaml
```diff
@@ -337,12 +266,20 @@ jobs:
...
+                run: "(cd tests/Application && bin/console doctrine:database:create)"
+
+            -   name: "Create database schema"
+                run: "(cd tests/Application && bin/console doctrine:schema:create)"
 
             -   name: "Collect code coverage with pcov and phpunit/phpunit"
                 run: "vendor/bin/phpunit --coverage-clover=.build/logs/clover.xml"
 
             -   name: "Send code coverage report to Codecov.io"
+                uses: "codecov/codecov-action@v4"
```
loevgaard: Use the newest action and I don't think you need to explicitly give it the files

.gitignore
loevgaard: Done in 6b78b5e — removed tests/Application/public/router.php and its .gitignore entry. Running the server with symfony serve is already documented in CLAUDE.md.

tests/Application/config/services_test.yaml
loevgaard: Done in 6b78b5e — removed. The functional tests use the test service container (self::getContainer()) to reach private services, so the public: true default is no longer needed; all 48 tests still pass.

tests/Application/config/packages/webpack_encore.yaml
loevgaard: Done in 6b78b5e — removed strict_mode: false. Verified the admin grid and shop pages render without any Encore entry errors.

src/Resources/translations/validators.en.yml
loevgaard: Done in 6b78b5e — removed the custom max_length messages here and the matching maxMessage options on the Length constraints (GiftCard.code and GiftCardDesignTranslation.name), so the Symfony default message is used. Kept the domain-specific and friendlier not_blank messages.

src/Resources/config/prepend/winzou_state_machine.yaml
loevgaard: Done in 527838e — prepend() now builds the configuration (winzou_state_machine, liip_imagine, sylius_mailer, sylius_ui, sylius_grid) as PHP arrays directly in the extension, and the prepend/*.yaml files are removed. Verified identical output via container lint and debug:config.

.github/workflows/build.yaml
loevgaard: Done in 6b78b5e — bumped to codecov-action@v5 and removed the explicit files input.

.github/workflows/build.yaml
```diff
@@ -89,26 +86,24 @@ jobs:
...
                     php-version: "${{ matrix.php-version }}"
-                    tools: "composer-require-checker, composer-unused, flex"
+                    tools: "flex"
 
             -   name: "Remove require-dev section in composer.json"
                 run: "composer config --unset require-dev"
-            
+
+            -   name: "Add composer-dependency-analyser"
+                run: "composer require --no-update --no-plugins --no-scripts --dev shipmonk/composer-dependency-analyser"
```
loevgaard: Instead of this, add the library using the `tools` in the setup-php action

src/Resources/config/services/pdf.xml
```diff
@@ -0,0 +1,21 @@
...
+            <argument>@SetonoSyliusGiftCardPlugin/shop/gift_card/pdf.html.twig</argument>
+            <argument>%setono_sylius_gift_card.pdf.page_size%</argument>
+            <argument>%sylius_core.public_dir%</argument>
+        </service>
+
+        <service id="Setono\SyliusGiftCardPlugin\Pdf\GiftCardPdfGeneratorInterface"
+                 alias="Setono\SyliusGiftCardPlugin\Pdf\DompdfGiftCardPdfGenerator"/>
+
+        <!-- kept as a convenience seam referenced by the mailer and controllers -->
+        <service id="setono_sylius_gift_card.pdf.generator" alias="Setono\SyliusGiftCardPlugin\Pdf\DompdfGiftCardPdfGenerator"/>
```
loevgaard: I don't think we should have this? Why should we?

src/Resources/config/doctrine/model/GiftCardTransaction.orm.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<doctrine-mapping xmlns:gedmo="http://gediminasm.org/schemas/orm/doctrine-extensions-mapping"
+                  xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping">
+    <mapped-superclass name="Setono\SyliusGiftCardPlugin\Model\GiftCardTransaction"
+                       table="setono_sylius_gift_card__gift_card_transaction">
+        <unique-constraints>
+            <unique-constraint name="setono_sylius_gift_card__transaction_idempotency" columns="idempotency_key"/>
+        </unique-constraints>
```
loevgaard: Instead of this, just add `unique=true` on the field

src/Resolver/GiftCardAwareDefaultPaymentMethodResolver.php
```diff
@@ -0,0 +1,51 @@
...
+        if ($payment instanceof PaymentInterface) {
+            $channel = $payment->getOrder()?->getChannel();
+            if ($channel instanceof ChannelInterface) {
+                foreach ($this->paymentMethodRepository->findEnabledForChannel($channel) as $candidate) {
+                    if ($candidate->getCode() !== $this->paymentMethodCode) {
+                        return $candidate;
+                    }
+                }
+            }
+        }
```
loevgaard, line 47: Do we need this block?

src/DependencyInjection/Compiler/AddAdjustmentsToOrderAdjustmentClearerPass.php
```diff
@@ -7,23 +7,27 @@
...
 use Symfony\Component\DependencyInjection\ContainerBuilder;
-use Webmozart\Assert\Assert;
 
+/**
+ * Registers the gift card adjustment type with Sylius' order adjustments clearer so that gift card
+ * adjustments are cleared and recomputed on every order processing run
+ */
 final class AddAdjustmentsToOrderAdjustmentClearerPass implements CompilerPassInterface
 {
+    private const PARAMETER = 'sylius.order_processing.adjustment_clearing_types';
```
loevgaard: Please check when this parameter was added to Sylius to make sure our plugin works in all versions we support

src/DependencyInjection/Compiler/ValidateAddToCartCommandClassPass.php
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * The plugin decorates the core add to cart command factory so the created command carries gift card information.
+ * If an application customizes the add to cart command it must stay compatible with that contract, so we fail fast
+ * (with an actionable message) when the configured command class does not implement our interface, instead of letting
+ * it surface later as an opaque "form view data is the wrong type" error
+ */
+final class ValidateAddToCartCommandClassPass implements CompilerPassInterface
+{
+    private const PARAMETER = 'setono_sylius_gift_card.order.model.add_to_cart_command.class';
```
loevgaard, line 20: This is not a core parameter. Is this a bug?

src/DependencyInjection/SetonoSyliusGiftCardExtension.php
```diff
@@ -4,64 +4,370 @@
...
+     * Prepends configuration for other bundles so the host application
+     * does not have to import any configuration files manually
+     */
+    public function prepend(ContainerBuilder $container): void
+    {
+        $operator = '@' . OrderGiftCardOperator::class;
+
+        $configuration = [
+            'winzou_state_machine' => [
+                'sylius_order_checkout' => [
```
loevgaard, line 65: Do we support both winzou and Symfony workflow state machines?

src/DependencyInjection/SetonoSyliusGiftCardExtension.php
```diff
@@ -4,64 +4,370 @@
...
+                                ],
+                            ],
+                        ],
+                    ],
+                ],
+            ],
+        ];
+
+        foreach ($configuration as $extension => $config) {
+            if ($container->hasExtension($extension)) {
```
loevgaard: Remove this if

src/Factory/GiftCardFactory.php
```diff
@@ -44,74 +36,32 @@ public function createForChannel(ChannelInterface $channel): GiftCardInterface
...
+        $baseCurrency = $channel->getBaseCurrency();
+        if (null !== $baseCurrency && null !== $baseCurrency->getCode()) {
+            $giftCard->setCurrencyCode($baseCurrency->getCode());
         }
 
         return $giftCard;
     }
 
-    public function createForChannelFromAdmin(ChannelInterface $channel): GiftCardInterface
+    public function createExample(ChannelInterface $channel, ?GiftCardDesignInterface $design = null): GiftCardInterface
```
loevgaard: What do we need this method for?

.github/workflows/build.yaml
loevgaard: Done in 2acb41e — set up as `tools: "flex, shipmonk/composer-dependency-analyser"` and invoked as `composer-dependency-analyser`, so the separate `composer require` step is gone. Installing it globally is also more correct here: the job unsets `require-dev` immediately before, and the analyser was the one thing being added back into the section it had just removed.

src/Menu/AdminMenuListener.php
```diff
@@ -27,5 +27,19 @@ private function addCatalogChild(ItemInterface $menu): void
...
+            ->setLabel('setono_sylius_gift_card.ui.gift_card_designs')
+            ->setLabelAttribute('icon', 'palette')
+        ;
+        $item
+            ->addChild('gift_card_balance', [
+                'route' => 'setono_sylius_gift_card_admin_gift_card_balance',
+            ])
+            ->setLabel('setono_sylius_gift_card.ui.outstanding_balance')
+            ->setLabelAttribute('icon', 'balance scale')
+        ;
```
loevgaard: These should probably be actions inside the main gift cards page, so we don't pollute the admin menu

src/Form/Type/AdjustGiftCardBalanceType.php
```diff
@@ -0,0 +1,54 @@
...
+use Sylius\Bundle\MoneyBundle\Form\Type\MoneyType;
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\TextareaType;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+use Symfony\Component\Validator\Constraints\NotBlank;
+use Symfony\Component\Validator\Constraints\NotEqualTo;
+use Webmozart\Assert\Assert;
+
+final class AdjustGiftCardBalanceType extends AbstractType
```
loevgaard, line 16: Should this have a data class? Maybe add this to CLAUDE.md that types preferably should have a data class?

src/Form/Type/GiftCardDesignTranslationType.php
```diff
@@ -0,0 +1,30 @@
...
+use Symfony\Component\Form\FormBuilderInterface;
+
+final class GiftCardDesignTranslationType extends AbstractResourceType
+{
+    public function buildForm(FormBuilderInterface $builder, array $options): void
+    {
+        // NotBlank lives on the GiftCardDesignTranslation entity (validation/GiftCardDesignTranslation.xml) rather than
+        // here: a form-level constraint would be applied to every rendered locale, forcing the name to be filled in all
+        // languages. On the entity it is only validated for the translations kept by ResourceTranslationsType (the
+        // default locale), so only the default locale name is required.
```
loevgaard, line 18: Remove this comment

src/Resources/config/services/pdf.xml
loevgaard: Done in 0b7d81c — removed. You are right, it was redundant: the `GiftCardPdfGeneratorInterface` alias right above it already provides an id to reference, so the mailer and the two controllers now use that. The "convenience seam" justification in the comment was circular — those three only referenced it because it existed.

src/Resources/config/doctrine/model/GiftCardTransaction.orm.xml
loevgaard: Done in 0b7d81c — replaced the `<unique-constraints>` block with `unique="true"` on the field. Verified the generated schema still has a unique index on `idempotency_key`; the only difference is the name, which goes from the explicit `setono_sylius_gift_card__transaction_idempotency` to Doctrine's generated `UNIQ_E4FE9367FD1C147`. No migration concern since 1.x is unreleased.

src/DependencyInjection/SetonoSyliusGiftCardExtension.php
loevgaard: Done in 0b7d81c — removed. I checked every extension prepended here (`winzou_state_machine`, `liip_imagine`, `sylius_mailer`, `sylius_ui`, `sylius_grid`) and each comes from a bundle that `sylius/core-bundle` hard-requires, so `hasExtension()` could never return false on a supported Sylius. It could only ever have hidden a misconfigured application.

src/DependencyInjection/Compiler/AddAdjustmentsToOrderAdjustmentClearerPass.php
loevgaard: Good catch — this was a real bug, fixed in 0b7d81c.

`sylius.order_processing.adjustment_clearing_types` was introduced in **Sylius 1.14.2** and never backported to 1.13 (checked up to the latest tag, v1.13.16). We support `~1.13.0 || ~1.14.0`, so on all of 1.13.x plus 1.14.0/1.14.1 the parameter is absent, `hasParameter()` returned false and the pass silently did nothing.

That mattered because `GiftCardAdjustmentProcessor` only ever *added* adjustments and delegated clearing entirely to the core clearer — so on those versions every order processing run appended another negative `order_gift_card` adjustment and the total kept dropping. CI did not catch it because no test processes an order twice in adjustment mode.

Rather than raise the Sylius floor, the processor now clears its own adjustments, which works on every supported version. Clearing happens before the early return, so adjustments no longer survive removing the last gift card either (previously only the core clearer handled that, i.e. only on >= 1.14.2). The compiler pass is now redundant and removed. The core clearer runs at priority 60 and this processor at 5, so it always runs first and cannot wipe what we just added. Added `GiftCardAdjustmentProcessorTest` covering both cases — I confirmed both tests fail if the clearing line is removed.

src/DependencyInjection/Compiler/ValidateAddToCartCommandClassPass.php
loevgaard: You are right that it is not a core parameter, but it is not a bug — it is our own parameter, declared in `services/order.xml:7` and defaulting to our `AddToCartCommand`. Sylius has no equivalent: `AddToCartCommandFactory` hardcodes `new AddToCartCommand(...)`, so there is nothing in core to hook.

We read the parameter in two places (the factory decorator in `order.xml:23` and `form.xml:42`), so it is a real extension point — the pass only fires if an application repoints it at a class that does not implement `AddToCartCommandInterface`, turning an opaque "form view data is the wrong type" into an actionable message.

Two things are fair to criticise though: `hasParameter()` can never be false since we always declare it, and the whole pass only guards against someone misconfiguring our own parameter. Happy to delete it if you consider that over-engineering — say the word.

src/Resolver/GiftCardAwareDefaultPaymentMethodResolver.php
loevgaard: Yes, it is what makes the class useful rather than merely restrictive.

`getDefaultPaymentMethod()` is called when Sylius creates a payment. The decorated resolver returns the first enabled method for the channel, which can be the gift card method (it is a real, enabled payment method in `payment` redemption mode). Without this block we would go straight to `UnresolvedDefaultPaymentMethodException` and break checkout for that channel; with it we fall through to the next enabled non-gift-card method. Deleting it would turn the class from "never default to the gift card method" into "break checkout whenever the gift card method sorts first".

It is untested though — there are no tests touching this resolver at all. I can add unit coverage for the three paths (decorated returns a normal method / returns the gift card method with a fallback available / no fallback available) if you want.

src/Factory/GiftCardFactory.php
loevgaard: It backs the admin "preview design as PDF" action — `PreviewGiftCardDesignPdfAction:40` builds a throwaway gift card so a design can be rendered without an order existing. It is the only caller.

That said, your instinct is reasonable: it puts demo data (a hardcoded 5000 amount and an English `Happy birthday!` message) on the factory interface, where it is visible to every consumer and not translatable. It would sit better inline in the preview action, leaving `GiftCardFactoryInterface` with just `createNew()` / `createForChannel()`. Want me to move it?

src/DependencyInjection/SetonoSyliusGiftCardExtension.php
loevgaard: No — we support winzou only, and this is a genuine gap worth deciding on.

We prepend callbacks under `winzou_state_machine` exclusively (reconcile on checkout complete, enable/send on pay, commit/rollback/disable on order transitions). Sylius 1.14 lets an application choose its adapter via `sylius_core.state_machine.default_adapter`, which defaults to `winzou_state_machine` but accepts `symfony_workflow`. On such an application none of our callbacks fire, so gift cards are never reconciled, enabled or emailed — silently.

The plugin is also inconsistent here: `PaymentRedemptionMethod` injects `Sylius\Abstraction\StateMachine\StateMachineInterface`, i.e. we already use the abstraction to *apply* transitions, while only registering *callbacks* for winzou. The abstraction covers applying transitions, not registering callbacks, so supporting both means also prepending equivalent `framework.workflows` listeners.

Since winzou is still the default in both 1.13 and 1.14, I have left this as-is rather than guess. Options as I see them: (a) leave winzou-only and document it, (b) also prepend Symfony Workflow callbacks, or (c) detect the configured adapter and fail loudly if it is not winzou. Which do you want?

src/Menu/AdminMenuListener.php
loevgaard: Agreed that three top-level entries is a lot for one plugin. Gift cards and designs are both resources with their own grids, so they are conventional menu items; the balance page is the one that is really a report *about* gift cards rather than a resource of its own.

My suggestion would be to keep `gift_cards` in the menu and move the other two onto the gift cards index page — designs as a grid action, balance as a link/button in the page header. That leaves one entry instead of three.

This is a visible admin UX change and needs Playwright verification per CLAUDE.md, so I have not made it yet. Confirm the split you want (keep just gift cards, or keep gift cards + designs) and I will do it.

src/DependencyInjection/SetonoSyliusGiftCardExtension.php
loevgaard: Done in bad8622 — we now support both.

Added six Symfony Workflow listeners under `EventListener/Workflow/`, matching the prepended winzou callbacks one to one and registered on `workflow.<graph>.completed.<transition>`, exactly as Sylius registers its own listeners. The graph and transition names are identical to the winzou ones (`sylius_order_checkout`/`complete`, `sylius_order_payment`/`pay`, `sylius_order`/`create`+`cancel`), so the two configurations line up directly.

Registering both sets is safe: Sylius applies a transition through a single configured adapter, so only that adapter emits events and the work is never done twice. Priorities preserve the ordering winzou got from declaration order — enable before send, rollback before disable.

Two things worth recording from checking this:
- `sylius_core.state_machine.default_adapter` exists in **both** 1.13 and 1.14, and both ship the `framework.workflows` definitions, so the gap applied to every version we support, not just 1.14.
- `symfony/workflow` is now referenced from `src`, so it is declared explicitly rather than leaning on `sylius/core-bundle` for it.

Verified all six are registered on the right events with the right priorities via `debug:event-dispatcher`, and added unit coverage for the listeners. I did **not** add an end-to-end test under the `symfony_workflow` adapter — that needs a second test environment plus an order fixture (gift card product, variant, items, units) that the suite has no helper for today. Worth doing separately if you want that covered.

src/Menu/AdminMenuListener.php
loevgaard: Done in 4afb5bc — gift cards keeps its menu entry, designs and the balance report are now actions on the gift cards index.

Two things surfaced while verifying it in the admin:

- With the full labels the button row **overflowed its container and clipped the last button**, even at 1680px wide, so the two new actions use short labels (`Designs`, `Balance`).
- The designs action could not keep the `palette` icon — **that name does not exist in Semantic UI**, so it silently rendered no icon at all. That was already true of the old menu entry. It uses `paint brush` now.

Both destinations would have become dead ends once out of the menu, so the designs grid gets an action leading back to gift cards, and the balance report gets gift cards in its breadcrumbs.

Verified with Playwright: menu down to one entry, all four buttons fitting, both links resolving, and back-navigation working from each destination.

One thing to check: I added `ui.designs` to the Danish and French catalogs (`Design` / `Modèles`) — worth a native eye. Note those two catalogs are already well out of sync with `en` from before this PR; they carry keys the rewrite dropped and are missing many it added.

src/DependencyInjection/Compiler/ValidateAddToCartCommandClassPass.php
loevgaard: Closing the loop on my offer to delete this: **keeping it** — and I tested it rather than reasoning about it.

Pointing the parameter at `stdClass` in the test application:

- **with the pass**, the container build fails immediately with the actionable message naming the parameter;
- **without it**, `cache:clear` reports `[OK]` and the container builds fine. The misconfiguration is just deferred to runtime.

Both consumers fail in the add-to-cart path — the shop's critical path. `AddToCartCommandFactory` does `new $this->className(...)` behind an `AddToCartCommandInterface` return type (TypeError, or ArgumentCountError if the constructor differs), and `AddToCartTypeExtension` sets `data_class` to it, which produces the opaque "form view data is the wrong type" error.

It also guards a real extension point rather than a hypothetical one: `AddToCartCommandFactory`'s idempotency check exists specifically so an application can extend our command and repoint this parameter. My earlier framing of it as merely guarding our own parameter undersold that.

The smaller criticism does stand though: the `hasParameter()` early return is effectively unreachable, since we always declare the parameter, and an application that unset it would fail the build anyway on the two services interpolating it. Happy to drop just that guard so the pass always validates.

loevgaard: Follow-up in 5e7e1ed: kept the pass, dropped the `hasParameter()` guard so it always validates.

src/Resolver/GiftCardAwareDefaultPaymentMethodResolver.php
loevgaard: Covered in 5e7e1ed — `GiftCardAwareDefaultPaymentMethodResolverTest` takes the four paths: the decorated resolver returning a normal method, returning the gift card method with a fallback available, the gift card method being the only enabled one, and a payment with no order. Confirmed they fail when the fallback block is removed, so they actually pin the behaviour rather than just executing it.

src/Factory/GiftCardFactory.php
loevgaard: Moved in 5e7e1ed — the construction now lives in `PreviewGiftCardDesignPdfAction`, its only caller, and `GiftCardFactoryInterface` is down to `createNew()` / `createForChannel()`. The message also became a translation key (`ui.preview_message`) rather than a hardcoded English string, so the preview is no longer English-only.

src/Form/Type/AdjustGiftCardBalanceType.php
loevgaard: Done in ffd5d25 — it binds to `AdjustGiftCardBalanceCommand` now, following `AddGiftCardToOrderCommand`, and the constraints moved onto that class as attributes.

Worth noting what it was costing: the action had to carry `/** @var array{amount: int, reason: string} $data */` to make anything of `$form->getData()`, which is exactly the annotation a data class removes.

Also added the preference to CLAUDE.md as you suggested: form types should bind to a `data_class` rather than produce an array, with constraints on that class, so the rules travel with the data instead of with the one form that happens to produce it.

Verified in the browser rather than just by the type checker: submitting zero/blank still gets rejected by the moved constraints, and a valid +12.34 adjustment lands as a `manual` ledger row with its reason, leaving the card reconciling at 7500 + 1234 = 8734.

src/Form/Type/GiftCardDesignTranslationType.php
loevgaard: Removed in ffd5d25.

src/Controller/Action/Admin/AdjustGiftCardBalanceCommand.php
```diff
@@ -0,0 +1,38 @@
...
+declare(strict_types=1);
+
+namespace Setono\SyliusGiftCardPlugin\Controller\Action\Admin;
+
+use Symfony\Component\Validator\Constraints as Assert;
+
+final class AdjustGiftCardBalanceCommand
+{
+    /** Minor units, positive to increase the balance and negative to decrease it */
+    #[Assert\NotNull(groups: ['setono_sylius_gift_card'])]
```
loevgaard: Shouldn't this be done i XML according to conventions?

src/Controller/Action/Admin/CreateGiftCardProductAction.php
```diff
@@ -0,0 +1,55 @@
...
+        ManagerRegistry $managerRegistry,
+        private readonly UrlGeneratorInterface $urlGenerator,
+    ) {
+        $this->managerRegistry = $managerRegistry;
+    }
+
+    public function __invoke(Request $request): Response
+    {
+        /** @var ProductInterface $product */
+        $product = $this->productFactory->create([
```
loevgaard: Fixture related code should not be used to produce live things imo

src/Controller/Action/Admin/GiftCardBalanceAction.php
```diff
@@ -0,0 +1,27 @@
...
+{
+    public function __construct(
+        private readonly GiftCardRepositoryInterface $giftCardRepository,
+        private readonly Environment $twig,
+    ) {
+    }
+
+    public function __invoke(): Response
+    {
+        $balances = $this->giftCardRepository->findBalance(new \DateTimeImmutable());
```
loevgaard: Maybe the default for `findBalance` should be 'now' so that we don't have to input the datetime if we just want 'now'

src/EventListener/Workflow/CommitRedemptionListener.php
```diff
@@ -0,0 +1,32 @@
...
+use Webmozart\Assert\Assert;
+
+/**
+ * Commits the gift card balance redeemed by an order when the order is placed
+ *
+ * This mirrors the equivalent winzou callback prepended in SetonoSyliusGiftCardExtension, so the plugin
+ * behaves the same whichever state machine adapter the application is configured with. Only the adapter
+ * actually applying the transition emits its events, so the two can never both run
+ */
+final class CommitRedemptionListener
```
loevgaard: Make this an event subscriber instead inside `src/EventSubscriber`

src/Controller/Action/Admin/AdjustGiftCardBalanceCommand.php
loevgaard: You're right — done in 87cff51.

Both command objects now declare their constraints in `src/Resources/config/validation/` (`AdjustGiftCardBalanceCommand.xml`, `AddGiftCardToOrderCommand.xml`), same as the models. I'd added the attributes on the argument that the rules should travel with the data class, but that argument is about *where the class lives*, not about the format — and XML is the format a host application can actually override by pointing at its own mapping file. An attribute is unreachable from the outside.

Verified they still fire: submitting the adjust balance form with amount 0 and an empty reason comes back with "This value should not be equal to 0." and "This value should not be blank.", exactly as before the move. I also corrected the CLAUDE.md convention line that said the opposite.

src/Controller/Action/Admin/CreateGiftCardProductAction.php
loevgaard: Agreed — fixed in 87cff51.

Extracted `Setono\SyliusGiftCardPlugin\Factory\GiftCardProductFactory` (+ interface). Building a gift card product — providing the shared `gift_card_delivery` option, one variant per delivery type with `shippingRequired` set accordingly, channel pricing — is domain logic and had no business living behind `AbstractExampleFactory`. `GiftCardProductExampleFactory` is now just the options resolver in front of the same factory, so fixtures and the admin action build identical products instead of the admin action borrowing the fixture path.

Checked both ways after a fixture reload from an empty database:
- fixtures produce the same rows as before (`gift_card-virtual` shipping_required=0, `gift_card-physical` =1, 5000 in FASHION_WEB)
- the admin action scaffolds a disabled gift card product with both variants and **reuses** the existing delivery option rather than duplicating it

Added a Playwright spec (`a gift card product can be scaffolded from the gift cards index`) since the action had no coverage at all; 25/25 green.

src/Controller/Action/Admin/GiftCardBalanceAction.php
loevgaard: Good call — done in 87cff51. `findBalance(?\DateTimeInterface $date = null)` now falls back to `new \DateTimeImmutable()`, so `GiftCardBalanceAction` just calls `findBalance()`. The parameter stays for the case where you deliberately want the outstanding balance at another point in time.

src/EventListener/Workflow/CommitRedemptionListener.php
loevgaard: Already done in 103d65a, which landed before you left this comment — the file is now `src/EventSubscriber/Workflow/CommitRedemptionSubscriber.php`. That commit converted every listener that could be a subscriber (all six workflow ones, `RecordGiftCardIssuanceSubscriber` and `AdminMenuSubscriber`), so the event name and priority sit in `getSubscribedEvents()` instead of a service tag.

Two stay listeners on purpose: `PendingGiftCardCleanupListener` reacts to Doctrine's `onFlush`, which is not a Symfony event, and Sylius' own `ImagesUploadListener`, which we cannot add the interface to.

src/Controller/Action/Admin/AdjustGiftCardBalanceCommand.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Setono\SyliusGiftCardPlugin\Controller\Action\Admin;
+
+final class AdjustGiftCardBalanceCommand
```
loevgaard, line 14: Should the gift card be a property on such a command?
Do we have validation for this command?

src/Controller/Action/Admin/CreateGiftCardProductAction.php
```diff
@@ -0,0 +1,53 @@
...
+        $this->managerRegistry = $managerRegistry;
+    }
+
+    public function __invoke(Request $request): Response
+    {
+        $product = $this->productFactory->create(
+            'gift_card_' . bin2hex(random_bytes(4)),
+            'Gift card',
+            enabled: false,
+        );
```
loevgaard: I would rather have a `do {} while` that checks for gift card codes and the first tested code should just be `gift_card`

src/DependencyInjection/Compiler/ValidateAddToCartCommandClassPass.php
loevgaard: Can you please double check that Sylius doesn't set a parameter for `AddToCartCommand`?

src/EventListener/PendingGiftCardCleanupListener.php
```diff
@@ -0,0 +1,46 @@
...
+use Setono\SyliusGiftCardPlugin\Model\OrderItemUnitInterface;
+
+/**
+ * Deletes pending (never paid) gift cards when their order item unit is removed, e.g. when a customer edits
+ * their cart or when Sylius prunes expired carts. Completed gift cards are never touched because the database
+ * foreign key is SET NULL and this listener only removes cards that are still pending
+ */
+final class PendingGiftCardCleanupListener
+{
+    public function onFlush(OnFlushEventArgs $args): void
```
loevgaard, line 19: Are we sure this should listen to onFlush and not preRemove? I think I remember Doctrine warns about listening to this event in their docs

src/Factory/GiftCardProductFactory.php
```diff
@@ -0,0 +1,200 @@
...
+        ProductInterface $product,
+        string $productCode,
+        ProductOptionValueInterface $optionValue,
+        GiftCardDeliveryType $deliveryType,
+        array $channels,
+        int $price,
+    ): ProductVariantInterface {
+        /** @var ProductVariantInterface $variant */
+        $variant = $this->productVariantFactory->createNew();
+        $variant->setCode(sprintf('%s-%s', $productCode, (string) $optionValue->getCode()));
```
loevgaard: The other place where we set a code we use `_`. Choose one: `-` or `_`

src/Factory/GiftCardProductFactoryInterface.php
```diff
@@ -0,0 +1,29 @@
...
+
+declare(strict_types=1);
+
+namespace Setono\SyliusGiftCardPlugin\Factory;
+
+use Setono\SyliusGiftCardPlugin\Model\GiftCardDeliveryType;
+use Setono\SyliusGiftCardPlugin\Model\ProductInterface;
+use Sylius\Component\Core\Model\ChannelInterface;
+
+interface GiftCardProductFactoryInterface
```
loevgaard, line 11: This should not extend the default Sylius factory?

src/Validator/Constraints/GiftCardIsApplicable.php
```diff
@@ -0,0 +1,31 @@
...
+
+declare(strict_types=1);
+
+namespace Setono\SyliusGiftCardPlugin\Validator\Constraints;
+
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * @Annotation
+ */
```
loevgaard: Is this still supported in the versions we support?

src/Controller/Action/Admin/AdjustGiftCardBalanceCommand.php
loevgaard: Both — done in 8b140e7.

**Validation**: it exists, just not in the class any more — it moved to `src/Resources/config/validation/AdjustGiftCardBalanceCommand.xml` in the previous round, per your XML comment. `amount` is `NotNull` + `NotEqualTo 0`, `reason` is `NotBlank`.

**The gift card as a property**: yes, and it turned out to matter. It is now a constructor argument, so the command describes the whole intent rather than half of it — and, more importantly, validation can now judge the adjustment against the balance it applies to. Deducting more than the card holds used to sail through the form and hit the balance operator:

```php
Assert::greaterThanEq($newAmount, 0, 'A manual gift card adjustment cannot make the balance negative');
```

which is a 500, for what is ordinary user error. There is a new `BalanceAdjustmentIsApplicable` class constraint that reports it on the `amount` field instead. Submitting -99999 against a $1,000.00 card now comes back 200 with "Deducting more than the gift card holds is not possible. The balance is $1,000.00." Covered by a Playwright spec.

That also uncovered a real bug — see the note on the translations in the PR.

src/Controller/Action/Admin/CreateGiftCardProductAction.php
loevgaard: Done in 8b140e7:

```php
private function provideCode(): string
{
    $suffix = 1;

    do {
        $code = 1 === $suffix ? self::CODE : sprintf('%s_%d', self::CODE, $suffix);
        ++$suffix;
    } while (null !== $this->productRepository->findOneBy(['code' => $code]));

    return $code;
}
```

Verified against the test app, where the fixture already owns `gift_card`: two consecutive clicks produced `gift_card_2` and `gift_card_3`. On a shop with no gift card product yet the first one is plain `gift_card`, as you wanted.

src/DependencyInjection/Compiler/ValidateAddToCartCommandClassPass.php
loevgaard: Checked — it does not. `Sylius\Bundle\OrderBundle\Factory\AddToCartCommandFactory` is `final` and hardcodes the class:

```php
public function createWithCartAndCartItem(OrderInterface $cart, OrderItemInterface $cartItem): AddToCartCommandInterface
{
    return new AddToCartCommand($cart, $cartItem);
}
```

Grepping `vendor/sylius` for `add_to_cart_command` turns up exactly one hit, the service id `sylius.factory.add_to_cart_command` — no `.class` parameter, and `AddToCartCommand` is not a configurable resource. The only seam is decorating that factory, which is what we do, so `setono_sylius_gift_card.order.model.add_to_cart_command.class` stays ours. If Sylius ever adds one we should switch to it rather than keep a second.

src/EventListener/PendingGiftCardCleanupListener.php
loevgaard: Pushing back on this one, but your question was worth asking — it had no tests at all, so I wrote some first (`tests/Functional/PendingGiftCardCleanupTest.php` in 8b140e7).

What the docs actually say (ORM 2.20, which is what we run):

- **preRemove** — "There are no restrictions to what methods can be called inside the `preRemove` event, **except when the remove method itself was called during a flush operation**."
- **onFlush** — "`onFlush` is a very powerful event. It is called inside `EntityManager::flush()` after the changes to all the managed entities and their associations have been computed."

The onFlush warnings are about *creating* and *updating*: a new entity needs `$unitOfWork->computeChangeSet()` on top of `persist()`, and a modified one needs `recomputeSingleEntityChangeSet()`. We only delete, so neither applies — `remove()` during onFlush lands in `entityDeletions` and `executeDeletions` picks it up.

The preRemove exception, on the other hand, is exactly our case. The way a cart really loses a unit is not `$em->remove($unit)`: the customer removes a line, Sylius removes the item, and `units` is mapped `orphan-removal="true"` with `cascade-all`, so the removal *is* triggered during the flush — the situation the docs carve out. (Sylius' `ExpiredCartsRemover` calls `$orderManager->remove($expiredCart)` then flushes, so that path cascades the same way.)

I did try it: with the listener switched to `preRemove` all three tests pass today. But that is Doctrine's documented grey area working by luck, and onFlush is the event that is specified to do this, so I would rather keep it and now have the tests to catch it either way.

src/Factory/GiftCardProductFactory.php
loevgaard: Agreed, `_` — fixed in 8b140e7. Variants are now `gift_card_virtual` / `gift_card_physical`, matching the product code `gift_card` and the option code `gift_card_delivery`. Reloaded the fixtures from an empty database to confirm; nothing referenced the dashed form.

src/Factory/GiftCardProductFactoryInterface.php
loevgaard: I would rather it did not, though tell me if you disagree and I will change it.

`Sylius\...\FactoryInterface` declares `createNew()`. A gift card product is not a gift card product until it has a code, the `gift_card_delivery` option and a variant per delivery type — so `createNew()` could only return a bare `Product` that fails to be the one thing the interface's name promises. Whoever called it would get something unusable.

The comparison with `GiftCardFactoryInterface` does not quite hold: that one *has* to implement `FactoryInterface` because it decorates `setono_sylius_gift_card.factory.gift_card`, the resource factory Sylius generates for the `gift_card` resource, and has to stand in for it. There is no `gift_card_product` resource — this factory decorates nothing and is only ever called directly, by the fixture factory and the admin action.

One thing I did fix while looking: `DEFAULT_PRICE` was on the implementation and the interface's default argument pointed at it, which is backwards. It lives on the interface now.

loevgaard:
Heads up on something that fell out of the balance-adjustment validation above: **almost every constraint message in the plugin was rendering as a raw translation key.**

Symfony resolves constraint messages in the `validators` domain, but nine of the ten lived in `messages.*.yml`. So a customer whose gift card was rejected in the shop saw literally

```
setono_sylius_gift_card.gift_card.not_enabled
```

instead of "This gift card is not active." — same for expired, no_balance, channel_mismatch, currency_mismatch, already_applied and both gift card amount limit messages. Only `gift_card_design.images.unique_type` was in the right place.

I noticed because my new message did exactly the same thing when I first put it in `messages.*.yml`. All ten now live in `validators.{en,da,fr}.yml` and the admin form renders "Deducting more than the gift card holds is not possible. The balance is $1,000.00." The Playwright spec asserts on the translated text, so a key leaking through fails the build from now on.

src/Validator/Constraints/GiftCardIsApplicable.php
loevgaard: Good catch — it is not, really. Removed from all four constraints in 3c6ed60.

`@Annotation` only means something to the Doctrine annotation reader, and Symfony 6.4 (the only Symfony we support) deprecates that path explicitly:

```
symfony/validator 6.4: "%s uses Doctrine Annotations to configure validation constraints,
which is deprecated. Use PHP attributes instead."
```

`framework.validation.enable_annotations` is deprecated in 6.4 too, in favour of `enable_attributes`, and Symfony 7 drops the reader altogether. Since we require PHP >= 8.1 the `#[\Attribute]` line underneath was already the only declaration anyone could use — and our own constraints are declared in XML anyway.

(`doctrine/annotations` 2.0.2 *is* still installed, via `sylius/resource-bundle`, so the docblock was not broken — just dead weight pointing at a deprecated mechanism.)

The one thing it changes: an application that wrote `@GiftCardIsApplicable` in a docblock rather than as an attribute would stop being validated. Deliberate, given we are pre-release and Symfony is removing it regardless.

PR #318, opened 2026-09-14
Validate the Referer before redirecting after apply/remove

### LINKED ISSUE #299: [Security]: Redirect after apply/remove follows an unvalidated Referer header

### What happened?
`src/Resolver/RedirectUrlResolver.php:32-38` redirects to the raw `Referer` header after `filter_var(..., FILTER_SANITIZE_URL)`, which strips characters but validates nothing, so the apply/remove actions redirect to whatever host the referer names. The `redirect` request-attribute branch (`:20-30`) is dead code: no route sets it. Exploitability is low (the browser controls its own Referer), but the standard hardening is trivial.

### Suggested fix
Accept the referer only when it starts with `$request->getSchemeAndHttpHost()` (or is a relative path), otherwise fall back to the default route. Remove the unused attribute branch or document it. Unit test with same-host, other-host and empty referer.

### Overlaps
None.

loevgaard:
## Summary

The apply/remove gift card actions redirect the customer back to wherever they came from by reading the `Referer` header and, after only `filter_var(..., FILTER_SANITIZE_URL)` (which strips characters but validates nothing), redirecting straight to it. Since the `Referer` is fully controlled by the client, this let the apply/remove flow redirect to an arbitrary host. `RedirectUrlResolver` now follows the `Referer` only when it stays on the current request's own host: a relative path (rejecting protocol-relative `//` and `/\` forms), or an absolute URL whose scheme, host and port match `$request->getSchemeAndHttpHost()`, with a boundary check so a value like `https://shop.example.com.evil.example.com` or `https://shop.example.com@evil.example.com` isn't mistaken for the real origin. Anything else, including `javascript:` URLs and a missing/empty referer, falls back to the default route as before. The `redirect` route default (`['route_name']` or `['route' => ..., 'parameters' => ...]`), which lets a host application pin the destination, is kept and still takes precedence over the Referer.

## Changes

- `src/Resolver/RedirectUrlResolver.php`: validate the `Referer` against the request's own scheme+host+port before using it as a redirect target; reject protocol-relative (`//`, `/\`) and cross-host/`javascript:` values; keep the `redirect` route default as the first choice of destination (restored after review, now covered by unit tests).
- `tests/Unit/Resolver/RedirectUrlResolverTest.php`: new unit tests (Prophecy for `UrlGeneratorInterface`) covering a same-host referer (path, query+fragment, bare origin, non-standard port, relative path), an off-site referer (another host, host-prefix confusion, userinfo confusion, port mismatch, scheme mismatch, protocol-relative forms, `javascript:`), and a missing/empty referer.

## Verification

- `vendor/bin/phpunit --testsuite unit` — pass (66 tests, 93 assertions)
- Regression check: stashed the `RedirectUrlResolver.php` fix and reran the new test file — 9 failures reproducing the open redirect (other host, host-prefix/userinfo/port/scheme confusion, both protocol-relative forms, `javascript:`, empty referer); restored the fix and it's green again.
- `DATABASE_URL=mysql://root@127.0.0.1/ssgc_referer_%kernel.environment%?serverVersion=5.7`, `doctrine:database:create --if-not-exists`, `vendor/bin/phpunit --testsuite functional` — pass (15 tests, 50 assertions)
- `composer analyse` (PHPStan max) — no errors
- `composer check-style` — no errors
- `vendor/bin/rector process --dry-run` — no changes needed
- Not run: `lint:yaml`/`lint:twig`/`lint:container`, `doctrine:schema:validate`, and Playwright — the change touches only PHP redirect-validation logic (no YAML/Twig/service/Doctrine/UI changes), so these don't apply.

## Notes for reviewers

- Kept the relative-path acceptance branch (`/en_US/cart/...`) since the issue's suggested fix explicitly calls it out, even though browsers always send an absolute URL in `Referer`; it's defense-in-depth for any caller that constructs a `Request` with a relative referer directly (as the new unit tests do).
- No other in-flight PRs appear to touch `RedirectUrlResolver.php` or the two controller actions that call it (`AddGiftCardToOrderAction`, `RemoveGiftCardFromOrderAction`).


src/Resolver/RedirectUrlResolver.php
```diff
@@ -17,27 +22,30 @@ public function __construct(private readonly UrlGeneratorInterface $router)
 
     public function getUrlToRedirectTo(Request $request, string $defaultRoute): string
     {
-        /** @var mixed $redirect */
-        $redirect = $request->attributes->get('redirect');
```
loevgaard, line 21: why was this functionality removed?

loevgaard: Good catch, that branch is a public extension point: a host can give the plugin's routes a `redirect` default (`[route]` or `{route, parameters}`) to pin where the apply/remove actions return. It was dropped only because nothing inside the plugin or the test app sets it. Restored in 5863961, still ahead of the Referer check, with unit tests for both forms and for a malformed attribute falling through to the Referer.

PR #322, opened 2026-09-14
Disable Dompdf remote resource loading and raise the dompdf version floor

### LINKED ISSUE #298: [Security]: Disable Dompdf remote resource loading and raise the dompdf version floor

### What happened?
- `src/Pdf/DompdfGiftCardPdfGenerator.php:34` sets `isRemoteEnabled` to `true`, but the generator only ever feeds Dompdf absolute local paths under the public dir (`:57-67`) with `chroot` set (`:36`). Remote loading is unnecessary and widens the attack surface (Dompdf's remote font/SVG CVEs required it). User content (`customMessage`) is Twig-escaped so this is not exploitable today, but defence in depth is cheap.
- `composer.json` requires `dompdf/dompdf: ^2.0 || ^3.0`; the lowest-deps CI job installs 2.0.0, and 2.0.0–2.0.3 have published advisories (CVE-2022-41343, CVE-2023-23924). The lock currently resolves 3.1.6.

### Suggested fix
- `isRemoteEnabled` → `false` (make it an opt-in constructor argument if a host renders remote images).
- Raise the floor to a patched 2.0.x (`^2.0.4 || ^3.0`) and let `composer audit` run in CI without `block: false` for this package.

### Acceptance criteria
`GiftCardPdfGeneratorTest` still renders front/back images with remote loading disabled; lowest-deps CI job green.

loevgaard:
## Summary

`DompdfGiftCardPdfGenerator` set `isRemoteEnabled` to `true`, but the generator only ever feeds Dompdf absolute local paths under the chrooted public dir (front/back design images), so remote loading bought no functionality while unnecessarily widening the attack surface that Dompdf's remote font/SVG CVEs relied on. `isRemoteEnabled` is now an opt-in constructor argument (`$remoteEnabled`), defaulted to `false` and wired explicitly to `false` in `pdf.xml`, so a host that genuinely needs remote images (e.g. served from a CDN) can still turn it on. Separately, `composer.json` allowed `dompdf/dompdf: ^2.0 || ^3.0`, so the lowest-deps CI job could resolve dompdf 2.0.0, which has published advisories for CVE-2022-41343 and CVE-2023-23924 (and its bypass, CVE-2023-24813). The floor is raised to `^2.0.4`, the first 2.0.x release fixing all three.

## Changes

- `src/Pdf/DompdfGiftCardPdfGenerator.php`: added a `bool $remoteEnabled = false` constructor argument and used it for `Options::set('isRemoteEnabled', ...)` instead of a hardcoded `true`. Chroot and the DejaVu default font are unchanged.
- `src/Resources/config/services/pdf.xml`: wired `$remoteEnabled` to `false` explicitly.
- `composer.json`: raised the `dompdf/dompdf` floor from `^2.0 || ^3.0` to `^2.0.4 || ^3.0`.
- `tests/Functional/GiftCardPdfGeneratorTest.php`: added `it_embeds_the_design_images_with_remote_loading_disabled`, which uploads real front/back PNGs through Sylius' image uploader (the same path the admin form and fixtures use), attaches them to a design, and asserts the generated PDF still has two pages and embeds both raster images as XObjects (an image Dompdf refuses to load is instead drawn as a vector "broken image" placeholder with no image XObject). A `tearDown()` removes the uploaded test images again.

## Verification

- `composer validate --strict` — pass
- `composer normalize --dry-run` — pass, already normalized
- `vendor/bin/phpunit --testsuite unit` — pass (51 tests, 73 assertions)
- `vendor/bin/phpunit --testsuite functional` (own DB `ssgc_dompdf`) — pass (16 tests, 64 assertions), including the new test
- `composer analyse` (PHPStan max) — pass, no errors
- `composer check-style` — pass
- `vendor/bin/rector process --dry-run` — pass, no errors
- `(cd tests/Application && bin/console lint:yaml ../../src/Resources && bin/console lint:twig ../../src/Resources && bin/console lint:container)` — pass
- Confirmed the new test also passes against the unmodified generator (local image loading is unaffected by `isRemoteEnabled`, since the generator never gave Dompdf anything but local absolute paths) — it's coverage for the acceptance criterion ("still renders front/back images with remote loading disabled"), not a regression test for a behavioural bug, so it isn't expected to fail on the old code; the security improvement itself (closing remote loading) isn't practically testable without contacting a real network host from a functional test.
- Doctrine mapping / UI: not applicable — no entity or template/JS/CSS changes in this PR.

## Notes for reviewers

- No CI workflow currently runs `composer audit`, so there's nothing to change there; the issue's mention of removing a `block: false` audit override doesn't apply to this repo as-is.
- `composer.lock` is gitignored, so this PR doesn't force a dompdf upgrade for anyone already on `vendor` (currently 3.1.6 locally); the floor change only affects fresh/lowest-deps installs.


src/Resources/config/services/pdf.xml
```diff
@@ -10,6 +10,7 @@
             <argument>@SetonoSyliusGiftCardPlugin/shop/gift_card/pdf.html.twig</argument>
             <argument>%setono_sylius_gift_card.pdf.page_size%</argument>
             <argument>%sylius_core.public_dir%</argument>
+            <argument key="$remoteEnabled">false</argument>
```
loevgaard: Make the $mediaDir an argument here also

composer.json
```diff
@@ -16,7 +16,7 @@
         "doctrine/collections": "^1.8 || ^2.0",
         "doctrine/orm": "^2.14",
         "doctrine/persistence": "^2.5 || ^3.0",
-        "dompdf/dompdf": "^2.0 || ^3.0",
+        "dompdf/dompdf": "^2.0.4 || ^3.0",
```
loevgaard, line 19: Do we know that our code works for both 2.0.4 and 3.0?

src/Resources/config/services/pdf.xml
loevgaard: Done in 8cf2669: `$mediaDir` is now passed explicitly (`media/image`) next to `$remoteEnabled`. Container lint and the PDF functional tests pass.

composer.json
loevgaard: Yes, both majors are exercised by CI on this PR (run 34816117464). The "Deps: lowest" functional jobs install dompdf v2.0.4 and the "Deps: highest" jobs install v3.1.6, and all four functional jobs run `GiftCardPdfGeneratorTest`, which renders a card with front and back design images through the real generator. The generator only uses API that is identical in 2.x and 3.x (`Options::set` for `isRemoteEnabled`, `defaultFont`, `chroot`; `setPaper`, `loadHtml`, `render`, `output`). Locally the vendor is 3.1.6 and the same tests pass.

src/Resources/config/services/pdf.xml
```diff
@@ -10,6 +10,8 @@
             <argument>@SetonoSyliusGiftCardPlugin/shop/gift_card/pdf.html.twig</argument>
             <argument>%setono_sylius_gift_card.pdf.page_size%</argument>
             <argument>%sylius_core.public_dir%</argument>
+            <argument key="$mediaDir">media/image</argument>
+            <argument key="$remoteEnabled">false</argument>
```
loevgaard: don't name arguments if they come in order

loevgaard: Fixed: both arguments are now positional (they follow the constructor order). Container lint and the PDF functional tests pass.

PR #323, opened 2026-09-14
Name every mapped column explicitly

### LINKED ISSUE #286: [Bug]: Physical column names depend on the host's Doctrine naming strategy; UPGRADE-1.0.md assumes underscores

### What happened?
Several mapped fields leave the column name implicit:
- `src/Resources/config/doctrine/model/GiftCard.orm.xml:13-17`: `initialAmount`, `currencyCode`, `deliveryType`, `customMessage`, `expiresAt`
- `src/Model/ProductTrait.php:11`: `giftCard`

The test app sets `naming_strategy: doctrine.orm.naming_strategy.underscore_number_aware` (`tests/Application/config/packages/doctrine.yaml:19-22`, its own comment says this is exactly to affect the plugin's implicit names). Sylius-Standard 1.13/1.14 sets **no** naming strategy, so a stock host gets `initialAmount`, `deliveryType`, `giftCard` columns, while `UPGRADE-1.0.md:37` tells it to backfill `delivery_type` / `initial_amount`. Migrations generated on two hosts differ.

### Suggested fix
- Add explicit `column="initial_amount"` etc. to every field in the XML mappings and `name: 'gift_card'` to the trait column (attribute and annotation), matching Sylius' underscore convention.
- Remove the naming strategy from the test app so CI runs the way Sylius-Standard does.
- Keep `UPGRADE-1.0.md` SQL consistent with the explicit names.

### Acceptance criteria
`doctrine:schema:validate` passes in the test app without a naming strategy; the generated migration on a stock host uses the underscore names in UPGRADE-1.0.md.

### Overlaps
Same trait files as #280 – do both in one PR or sequence them.

loevgaard:
Its parent #317 has been merged into 1.x, so this PR now targets 1.x directly.

## Summary

Several of the plugin's fields were mapped without a column name, which means the physical name is decided by the *host application's* `doctrine.orm.naming_strategy` rather than by the plugin. Sylius-Standard 1.13/1.14 configures no strategy, so a stock store ends up with `initialAmount`, `currencyCode`, `deliveryType`, `customMessage`, `expiresAt` and `giftCard` columns — while `UPGRADE-1.0.md` tells it to backfill `delivery_type` and `initial_amount`, and a store that does configure an underscore strategy gets a different schema again for the same code. The test application hid this by configuring `underscore_number_aware` itself.

Every column the plugin maps is now named explicitly, the test application configures no naming strategy (exactly like Sylius-Standard), and a functional test reads the physical names back off the class metadata so the regression cannot return unnoticed.

## Changes

- `src/Resources/config/doctrine/model/*.orm.xml`: every `<field>` now carries a `column` attribute — `initial_amount`, `currency_code`, `delivery_type`, `custom_message`, `expires_at` on `GiftCard`, plus the single-word fields on `GiftCard`, `GiftCardDesign`, `GiftCardDesignTranslation` and `GiftCardTransaction` so nothing is left to a strategy at all.
- `src/Model/ProductTrait.php`: `name: 'gift_card'` / `name="gift_card"` on the `giftCard` flag, in both the attribute and the annotation.
- `tests/Application/config/packages/doctrine.yaml`: removed `naming_strategy: doctrine.orm.naming_strategy.underscore_number_aware`, so the test app builds its schema the way a stock Sylius application does.
- `UPGRADE-1.0.md`: the backfill step now shows the actual SQL, and a new "Column names" subsection states that the names are strategy-independent and warns `0.12.x` stores that ran without a naming strategy that `doctrine:migrations:diff` renders the change as a drop-plus-add — with the `ALTER TABLE ... CHANGE` statements to use instead.

## Verification

Run from the worktree against a private database (`ssgc_column_names_*`) and port 8118.

- `vendor/bin/phpunit` — OK (89 tests, 169 assertions); `--testsuite unit` OK (51), `--testsuite functional` OK (38, including the 23 new mapping assertions).
- Regression check: with the mapping changes reverted and the test app without a naming strategy, the created schema has `initialAmount`, `deliveryType` and `giftCard` columns; with the fix the `information_schema` read-back shows `initial_amount`, `delivery_type` and `gift_card`. (A `ClassMetadata`-based test that pinned the names was dropped after review, since it only re-read the mapping file.)
- `doctrine:schema:drop --full-database` + `doctrine:schema:create` + `doctrine:schema:validate -vvv` — "[OK] The mapping files are correct."
- The created schema was read back from `information_schema`: `initial_amount`, `currency_code`, `delivery_type`, `custom_message`, `expires_at` on `setono_sylius_gift_card__gift_card` and `gift_card` on `sylius_product`, with no naming strategy configured.
- `composer analyse` — "[OK] No errors". `composer check-style` — "[OK] No errors found". `vendor/bin/rector process --dry-run` — "[OK] Rector is done!".
- `bin/console lint:yaml ../../src/Resources` (14 files) and `lint:yaml config` (44 files) — OK. `bin/console lint:container` — OK.
- Playwright, against `symfony serve` on port 8118 with freshly loaded fixtures on the new schema: `npx playwright test` — 26 passed (shop + admin). No spec was added: this change has no UI surface, but the suite was run because the physical schema changed under it. (The first run failed the auth setup on a cold dev-container cache; it passed on the immediate re-run.)

## Notes for reviewers

- `doctrine:schema:validate` still reports 56 schema diffs in the Database section, all of the `CHANGE x x DATETIME DEFAULT NULL` / `JSON NOT NULL` kind and mostly on untouched Sylius core tables (`sylius_product`, `sylius_address`, `sylius_promotion`, …). That is pre-existing DBAL comparator noise against MySQL, not something this PR introduces — none of the diffs renames a column, and the mapping section is clean.
- Single-word fields (`code`, `amount`, `enabled`, `version`, `position`, `type`, `reason`, `name`) do not change name under any common strategy, but they are named too so the mappings state the whole physical schema rather than half of it; the test asserts them for the same reason.
- `src/Model/ProductTrait.php` is also touched by #317 (the attribute mapping), which is why this is stacked on it rather than on `1.x`.


src/Resources/config/doctrine/model/GiftCard.orm.xml
```diff
@@ -7,15 +7,18 @@
...
-        <field name="enabled" type="boolean"/>
-        <field name="amount" type="integer"/>
-        <field name="initialAmount" type="integer"/>
-        <field name="currencyCode" length="3"/>
-        <field name="deliveryType" type="string" enum-type="Setono\SyliusGiftCardPlugin\Model\GiftCardDeliveryType"/>
-        <field name="customMessage" type="text" nullable="true"/>
-        <field name="expiresAt" type="datetime" nullable="true"/>
-        <field name="version" type="integer" version="true"/>
+        <!-- Every column is named explicitly: an implicit name is resolved by the *application's* naming strategy, so
+             two host applications would end up with different physical schemas for the same plugin -->
```
loevgaard: Remove this comment

tests/Functional/DoctrineColumnNamingTest.php
```diff
@@ -0,0 +1,128 @@
...
+
+/**
+ * The physical column name of a field mapped without an explicit column is decided by the *application's*
+ * Doctrine naming strategy, so two host applications would end up with different schemas for the same plugin and
+ * the SQL in UPGRADE-1.0.md would only be right for some of them.
+ *
+ * The test application deliberately configures no naming strategy (like Sylius-Standard), which makes Doctrine fall
+ * back to the field name verbatim. These assertions therefore only hold while every column is named explicitly.
+ */
+final class DoctrineColumnNamingTest extends KernelTestCase
```
loevgaard: Do we really need this test?

src/Resources/config/doctrine/model/GiftCard.orm.xml
loevgaard: Removed in c50e66c.

tests/Functional/DoctrineColumnNamingTest.php
loevgaard: No. It only re-read the mapping file through ClassMetadata, so it could not catch anything the mapping itself does not already state. Dropped in c50e66c; unit, functional, schema:validate, lint and the static checks are still green.

Issue #316, opened 2026-09-11
[Chore]: Translation cleanup: missing referenced key, unused keys, shadowed Sylius key, PHP-side English strings, default Symfony messages


loevgaard:
### Items
- **Referenced but missing**: `setono_sylius_gift_card.ui.gift_card_code_does_not_exist` (`src/Form/DataTransformer/GiftCardToCodeDataTransformer.php:56`; masked today because the form's `invalid_message` wins, while `ui.gift_card_code_is_invalid` exists unused).
- **Default Symfony messages shown to users**: `src/Resources/config/validation/AddGiftCardToOrderCommand.xml:9` `NotBlank` has no message ("This value should not be blank." while `validators.*.yml:6` defines an unused "Please enter code"); `AdjustGiftCardBalanceCommand.xml:12-23` likewise ("This value should not be equal to 0.").
- **Defined but never referenced** (messages): `form.gift_card.channel`, `form.gift_card_design.front_image`, `form.gift_card_design.back_image`, `ui.amount`, `ui.applied_gift_cards`, `ui.create_gift_card`, `ui.create_gift_card_design`, `ui.gift_card_channel_does_not_match_channel`, `ui.gift_card_code_is_invalid`, `ui.physical`, `ui.virtual`, `pdf.how_to_use`, `pdf.how_to_use_text`, `email.your_gift_card`; (validators): `add_gift_card_to_order_command.gift_card.is_expired`, `add_gift_card_to_order_command.gift_card.not_blank`, `channel_configuration.unique_channel_locale` (0.x leftover), `gift_card.amount.min`.
- **Shadowed Sylius key**: `sylius.form.locale.select` in `messages.{en,da,fr}.yml:103-106` overrides a key that exists in `SyliusLocaleBundle` ("Select a locale..."); the comment "Missing from Sylius for some reason" is wrong. Remove.
- **PHP-side English reaching customers in every locale**: product name `'Gift card'` (`CreateGiftCardProductAction.php:42`), option `'Delivery'` and values/variant names `ucfirst($deliveryType->value)` (`GiftCardProductFactory.php`), default design `'Classic'` (`GiftCardDesignProvider.php:54`), payment method name `'Gift card'` shown on order pages (`GiftCardPaymentMethodProvider.php:54`). At least the payment method should be named via the translator for every channel locale (`ui.gift_card`, `ui.virtual`, `ui.physical` exist).
[...]

Issue #287, opened 2026-09-11
[Bug]: Displayed code grouping is hardcoded to 4-4-4-4 slicing instead of GiftCardCodeNormalizer::format()


loevgaard:
_Context: plugin 1.x @ 090e4ec._

### What happened?
Displayed codes are grouped by hand-written slicing that assumes exactly 16 characters:
- `src/Resources/views/shop/gift_card/pdf.html.twig:2`: `code|slice(0,4) ~ '-' ~ code|slice(4,4) ~ '-' ~ code|slice(8,4) ~ '-' ~ code|slice(12)`.

But `code_length` is configurable 4..255 (`src/DependencyInjection/Configuration.php:36-41`) and admins can type any code in `GiftCardType` (`src/Form/Type/GiftCardType.php:42`), so `SUMMER26` renders as `SUMM-ER26--`. The cart (`src/Resources/views/shop/cart/_gift_cards.html.twig:21`), the emails and the admin show page display the raw code. `GiftCardCodeNormalizer::format()` (`src/Generator/GiftCardCodeNormalizer.php:16-25`) implements the grouping already and is never called.

Related: #94.

### Suggested fix
- Add a Twig filter (e.g. `setono_gift_card_format_code`) in `src/Twig/` backed by `GiftCardCodeNormalizerInterface::format()` and use it in the PDF (all three places), the cart table, both email templates and `admin/gift_card/show.html.twig`.
- Make `format()` robust for codes shorter than one group and for codes containing existing separators.

### Acceptance criteria
- Unit tests for the filter/`format()` with 8, 15, 16 and 20 character codes.
- Rendering the PDF for an admin code `SUMMER26` shows `SUMM-ER26`.

### Overlaps for parallel work
`pdf.html.twig` is also touched by #302 and #295; the input side is #283.

PR #326, opened 2026-09-14
Normalize gift card codes entered in the cart

### LINKED ISSUE #283: [Bug]: Codes entered in the cart are not normalized, so the dashed code printed on the PDF is rejected
_Context: plugin 1.x @ 090e4ec; Sylius 1.13 / 1.14._

### Severity
High (customers cannot redeem the code as printed).

### What happened?
The PDF prints the code as `XXXX-XXXX-XXXX-XXXX` (`src/Resources/views/shop/gift_card/pdf.html.twig:2`, shown on the front scrim, in the code frame and under the barcode). The only shop path for entering a code is the cart form, whose data transformer does an **exact** lookup with the raw input:

- `src/Form/DataTransformer/GiftCardToCodeDataTransformer.php:47-50` → `GiftCardRepository::findOneEnabledByCodeAndChannel($value, ...)` (`src/Doctrine/ORM/GiftCardRepository.php:35-45`, plain `findOneBy`).

`GiftCardCodeNormalizer::normalize()` (uppercase + strip non-alphanumerics) exists but is only called from the programmatic path `src/Applicator/GiftCardApplicator.php:81`, which the form never reaches because the transformer resolves the entity first. A customer copying `ABCD-EFGH-JKMN-PQRS` off the card (or typing with spaces / lowercase on a case-sensitive collation) gets "The gift card with this code does not exist".

Related: #94 asked for exactly this (display with dashes, strip on input).

### Suggested fix
- Inject `GiftCardCodeNormalizerInterface` into `GiftCardToCodeDataTransformer` and normalize before the lookup.
- Consider normalizing in `GiftCardRepository::findOneByCode()` / `findOneEnabledByCodeAndChannel()` too, so every caller benefits.
- Keep `RemoveGiftCardFromOrderAction`'s CSRF token id consistent: it uses the raw `{giftCard}` route parameter (`src/Controller/Action/RemoveGiftCardFromOrderAction.php:41-42`) while the template passes `giftCard.code`; after normalizing, build the token id from the normalized code on both sides.

### Acceptance criteria
- Unit test (Prophecy) for the transformer: `abcd-efgh-jkmn-pqrs`, ` ABCD EFGH JKMN PQRS `, and `abcdefghjkmnpqrs` all resolve the same card; unknown code still throws `TransformationFailedException`.
- Playwright: apply the seeded card as lowercase with separators and assert it is applied.

Display side is #287. Messages are touched by #316.

loevgaard:
## Summary

The PDF prints a gift card's code grouped by dashes (`XXXX-XXXX-XXXX-XXXX`), but the only shop path for entering a code — the cart form — did an exact lookup with the raw input. `GiftCardCodeNormalizer` already existed and was used by the programmatic redemption path (`GiftCardApplicator`), but the form's data transformer resolved the entity itself via `GiftCardRepository`, bypassing it entirely. A customer who copied the code off the card, or typed it with different casing, got "The gift card with this code does not exist". This wires the existing normalizer into the cart form's data transformer and the remove action, so the cart accepts a code in any of its equivalent renderings, and keeps the remove action's CSRF token id consistent with whatever form the code arrives in.

## Changes

- `GiftCardToCodeDataTransformer` now takes `GiftCardCodeNormalizerInterface` and normalizes the typed value before asking the repository for it.
- `RemoveGiftCardFromOrderAction` normalizes the `{giftCard}` route parameter before building the CSRF token id and before calling the applicator, so it stays consistent with whatever form the cart template passes as the code (today the stored canonical code; the display side, #287, may change what's shown without needing to touch this action).
- Playwright: added a case to `tests/Playwright/specs/shop/redemption.spec.js` applying the seeded card typed lowercase and grouped with dashes, asserting it is applied (via the stable remove-form action) and that it fully covers the order.
- Unit test (Prophecy) for the transformer covering the three equivalent renderings from the issue plus the unknown-code case.

## Verification

- `vendor/bin/phpunit --testsuite unit` — pass (55 tests)
- `vendor/bin/phpunit --testsuite functional` (own DB `ssgc_code_normalization`) — pass (19 tests)
- `composer analyse` — no errors
- `composer check-style` — no errors
- `vendor/bin/rector process --dry-run` — no changes needed
- `bin/console lint:yaml ../../src/Resources` — pass (14 files)
- `bin/console lint:twig ../../src/Resources` — pass (18 files)
- `bin/console lint:container` — pass
- Regression check: stashed the `src/` changes and reran the new unit + functional tests — both failed (transformer: `TransformationFailedException` on the dashed/lowercase/spaced inputs), confirming they exercise the bug. Restored the fix and reran — both pass.
- Playwright shop project against the served test app (own DB + port 8107): `npx playwright test --project=shop` — 9 passed, including the new normalization case and the two pre-existing redemption/removal specs.

## Notes for reviewers

- Left `src/Resources/views/shop/cart/_gift_cards.html.twig` untouched per instructions — #287 is already changing how the code is displayed there.
- Did not touch translation files per #316.
- The repository stays a plain lookup: both callers, the transformer and `GiftCardApplicator::resolveGiftCard()`, normalize before asking it (changed after review; an earlier revision normalized inside the repository and injected the normalizer through the DI extension).


src/DependencyInjection/SetonoSyliusGiftCardExtension.php
```diff
@@ -47,6 +50,15 @@ public function load(array $configs, ContainerBuilder $container): void
...
 
+        // The resource bundle builds the repository from the manager and the class metadata alone, so the
+        // normalizer its lookups by code rely on is appended here. A host application may swap in its own
+        // repository class, in which case only one still derived from ours is given the extra argument
+        $repository = $container->getDefinition('setono_sylius_gift_card.repository.gift_card');
+        $repositoryClass = $container->getParameterBag()->resolveValue($repository->getClass());
+        if (is_string($repositoryClass) && is_a($repositoryClass, GiftCardRepository::class, true)) {
+            $repository->addArgument(new Reference(GiftCardCodeNormalizerInterface::class));
+        }
+
```
loevgaard: This is so wrong. Either normalize the code before handing it to the repo or create a whole new 'data provider' service for fetching gift cards.

loevgaard: Agreed, that was the wrong place. Reverted in 0bbf8cf: the repository and the DI extension are back to what is on 1.x, the repository-level functional test is gone, and normalization happens only in the callers (the data transformer, and the remove action before it builds the CSRF token id and calls the applicator, which already normalized). Unit, functional, PHPStan, ECS, Rector and container lint are green.

PR #327, opened 2026-09-14
Resolve the admin design PDF preview channel from the design, not the shop context

### LINKED ISSUE #304: [Bug]: Admin design PDF preview resolves the channel from the shop channel context

### What happened?
`src/Controller/Action/Admin/PreviewGiftCardDesignPdfAction.php:42-43` calls `sylius.context.channel` from an **admin** request. That composite context resolves by request hostname or falls back to the single channel; on a multi-channel host whose admin runs on its own hostname it throws `ChannelNotFoundException` (500), and on a shared hostname it may pick a channel the design is not assigned to (wrong currency in the preview).

### Suggested fix
Use the design's own channel (`$design->getChannels()->first()`), fall back to the first channel from `sylius.repository.channel`, or accept an optional `channel` query parameter with a chooser on the design form.

### Acceptance criteria
`WebTestCase`/Playwright: preview works with two channels and no channel hostname matching the admin host.

loevgaard:
Update after review: the optional `?channel=` query parameter was dropped (the preview always uses the design's first channel, falling back to the shop's first channel), and the action now depends on Sylius' `ChannelRepositoryInterface` instead of the generic `RepositoryInterface`.

## Summary

`PreviewGiftCardDesignPdfAction` resolved its channel via `sylius.context.channel`, which resolves by request hostname. Called from an admin request, that context has nothing to fall back to on a multi-channel shop whose admin runs on its own hostname, throwing `ChannelNotFoundException` (500); on a shared hostname it could also pick a channel the design isn't assigned to, previewing the wrong currency. The action now resolves the channel from the design itself: an optional `channel` query parameter naming one of the design's channels, otherwise the design's first channel, falling back to the shop's first channel (`sylius.repository.channel`) only when the design has none yet. A requested channel the design isn't assigned to is refused with 404 rather than silently substituted.

## Changes

- `PreviewGiftCardDesignPdfAction`: replaced the `ChannelContextInterface` dependency with `RepositoryInterface<ChannelInterface>` (`sylius.repository.channel`), added a `Request $request` parameter, and added `resolveChannel()` implementing the design-first/query-param/repository-fallback/404 logic.
- `controller.xml`: updated the service's argument from `sylius.context.channel` to `sylius.repository.channel`.
- Added `tests/Unit/Controller/Action/Admin/PreviewGiftCardDesignPdfActionTest.php` (Prophecy): design's first channel is used; the `channel` query parameter selects among the design's channels; an empty `channel` parameter is treated as absent; a channel the design isn't assigned to is rejected with 404; a design with no channels falls back to the shop's first channel; no channel at all is a 404; an unknown design id is a 404.
- Added `tests/Functional/PreviewGiftCardDesignPdfActionTest.php` (`KernelTestCase`, following the existing `GiftCardFunctionalTestCase` convention): invokes the real, container-wired action against a real multi-channel database, proving the `sylius.repository.channel` DI wiring is correct and that the design's own channel is used regardless of any request hostname.
- Extended `tests/Playwright/specs/admin/gift-cards.spec.js` with a case that requests the preview PDF for one of the design's own channels (discovered from the edit form's channel checkboxes, not hardcoded) and asserts an unknown `channel` query value 404s.

## Verification

- `vendor/bin/phpunit --testsuite unit` — pass (58 tests, 88 assertions).
- `vendor/bin/phpunit --testsuite functional` — pass (19 tests, 56 assertions), against `ssgc_preview_channel_test` (MySQL 5.7).
- `composer analyse` (PHPStan max) — clean.
- `composer check-style` — clean.
- `vendor/bin/rector process --dry-run` — clean.
- `bin/console lint:yaml`, `lint:twig`, `lint:container` (services.xml changed) — clean.
- Regression check: stashed the `src/` changes and reran both new test files against the unmodified code — all fail (unit tests via a constructor `TypeError` since the old code took `ChannelContextInterface`; functional test via an `__invoke` signature `TypeError` since the old action didn't take a `Request`), then pass again after restoring the fix.
- Playwright: served `tests/Application` on port 8115 against a freshly seeded `ssgc_preview_channel` database, ran `npx playwright test --project=admin` — all 19 admin specs pass, including the two design-preview-PDF specs.

## Notes for reviewers

- While testing this in the shared multi-worktree setup, an unrelated `git stash` from another in-flight fix branch (`fix/code-normalization`) briefly landed in this worktree via a shared `refs/stash` collision (a `git stash pop` here picked up someone else's concurrently-pushed stash instead of my own). It was caught immediately via `git status`/`git diff` before anything was committed, backed up, and re-stashed for that branch to recover; this PR's diff only ever contains files relevant to #304. Flagging in case the `code-normalization` PR needs to double check its own working tree.
- The existing functional test suite has no precedent for `WebTestCase`/`KernelBrowser` (the project doesn't depend on `symfony/browser-kit`), so the new functional test follows the established `GiftCardFunctionalTestCase`/`KernelTestCase` pattern and invokes the container-wired action directly rather than through an HTTP client; the Playwright spec covers the real HTTP path end-to-end instead.


src/Controller/Action/Admin/PreviewGiftCardDesignPdfAction.php
```diff
@@ -8,39 +8,48 @@
...
 
+    /**
+     * @param RepositoryInterface<ChannelInterface> $channelRepository
+     */
     public function __construct(
         private readonly GiftCardDesignRepositoryInterface $designRepository,
         private readonly GiftCardFactoryInterface $giftCardFactory,
         private readonly GiftCardPdfGeneratorInterface $pdfGenerator,
-        private readonly ChannelContextInterface $channelContext,
+        private readonly RepositoryInterface $channelRepository,
```
loevgaard: Inject the `\Sylius\Component\Channel\Repository\ChannelRepositoryInterface` instead of the generic `RepositoryInterface`

src/Controller/Action/Admin/PreviewGiftCardDesignPdfAction.php
```diff
@@ -8,39 +8,48 @@
...
+    public function __invoke(Request $request, int $id): Response
     {
         $design = $this->designRepository->find($id);
         if (!$design instanceof GiftCardDesignInterface) {
             throw new NotFoundHttpException();
         }
 
-        $channel = $this->channelContext->getChannel();
-        Assert::isInstanceOf($channel, ChannelInterface::class);
+        $channelCode = $request->query->getString('channel');
```
loevgaard: Why do we need this?

src/Controller/Action/Admin/PreviewGiftCardDesignPdfAction.php
loevgaard: Done in b1efa3e: the constructor now takes `Sylius\Component\Channel\Repository\ChannelRepositoryInterface` (the `sylius.repository.channel` service already implements it, so the wiring is unchanged).

src/Controller/Action/Admin/PreviewGiftCardDesignPdfAction.php
loevgaard: We do not. The optional `?channel=` query parameter was an extra I added on top of the issue; nothing in the admin UI links to it. Removed in b1efa3e together with its unit, functional and Playwright cases: the preview now always renders for the design's first channel and falls back to the shop's first channel when the design has none. Unit, functional, PHPStan, ECS, Rector and container lint are green.

Issue #309, opened 2026-09-11
[Tests]: No form type tests exist although CLAUDE.md prescribes TypeTestCase


loevgaard:
### What happened?
CLAUDE.md says "Form type tests extend `Symfony\Component\Form\Test\TypeTestCase`", but `tests/Unit/Form` does not exist and no test references `TypeTestCase`. Nine form types, two extensions and one data transformer are untested:
`GiftCardInformationType`, `AddGiftCardToOrderType`, `AdjustGiftCardBalanceType`, `GiftCardType`, `GiftCardDesignType`, `GiftCardDesignImageType`, `GiftCardDesignTranslationType`, `CustomerAutocompleteChoiceType`, `Rule\HasNoGiftCardConfigurationType`, `AddToCartTypeExtension`, `ProductTypeExtension`, `GiftCardToCodeDataTransformer`.

### Suggested tests
- `GiftCardInformationType`: valid submission maps amount (minor units), message, design; empty amount → error not exception (see #282); message over the limit → error; needs Prophecy doubles for `ChannelContextInterface` and `GiftCardDesignProviderInterface` via `PreloadedExtension`.
- `AddToCartTypeExtension`: `giftCardInformation` child added only for gift card products; handler called on valid submit only.
- `AddGiftCardToOrderType` + transformer: normalized lookup (see #283), unknown code → `invalid_message`.
- `GiftCardType`: `amount` present on create and absent on edit; `initialAmount` set from `amount` on create; `currencyCode` behaviour after #293; expiry pinned to 23:59:59.
- `AdjustGiftCardBalanceType`: currency option required; command validation groups.
- `GiftCardDesignType`/`ImageType`: image collection add/remove; empty position (see #282).

### Overlaps
#282, #283, #293.

Issue #294, opened 2026-09-11
[Feature request]: Let admins choose a design and delivery type when issuing a gift card


loevgaard:
### Describe the proposed solution
The admin gift card form has neither a `design` nor a `deliveryType` field (`src/Form/Type/GiftCardType.php`), so every admin-issued card is `Virtual` with the framed default PDF. Add:
- `design`: `EntityType` over `GiftCardDesignRepositoryInterface::findEnabledByChannel()` for the chosen channel (the channel is only chosen on create, so add the field in the same `PRE_SET_DATA` branch, or filter by all channels on edit), optional, with thumbnails like the shop picker.
- `deliveryType`: `EnumType` over `GiftCardDeliveryType`, default `Virtual`.

### Additional context
Model already supports both (`GiftCard::setDesign()`, `setDeliveryType()`); the PDF generator picks images from the design. Admin show page displays both.

### Overlaps
`GiftCardType.php` also touched by #293 and #282.

Issue #293, opened 2026-09-11
[Bug]: The currency of a live gift card can be changed in admin, and it is never validated against the channel


loevgaard:
_Context: plugin 1.x @ 090e4ec._

### What happened?
`GiftCardType` adds `currencyCode` on **every** edit (`src/Form/Type/GiftCardType.php:110-129`, choices = all currencies) while `channel` (`:51-54`) and `amount` (`:59-65`) are deliberately locked after creation. Changing DKK to EUR on a redeemed card silently changes its value; `GiftCardTransaction` rows carry no currency of their own and `admin/gift_card/show.html.twig:118` formats them with the card's current currency. Nothing validates that the currency belongs to the card's channel (`src/Resources/config/validation/GiftCard.xml` has no such constraint), so an admin can issue a card the channel can never redeem (`GiftCardIsApplicableValidator.php:72-75` rejects it at the cart).

Lower priority in the same form: `customer` can be reassigned on a purchased card with no ledger trace.

### Suggested fix
- Add `currencyCode` only when `null === $giftCard->getId()` (or render it disabled on edit), mirroring `channel`.
- Add an XML constraint (custom `GiftCardCurrencyBelongsToChannel` or a callback) in `validation/GiftCard.xml` with group `setono_sylius_gift_card`, plus `Currency` on the code.
- Optional: write a `manual` ledger row (amount 0, reason) when the customer changes, or log it.

### Acceptance criteria
`TypeTestCase`: currency field absent/disabled on edit; validator test: currency not in channel → violation.

### Overlaps
`GiftCardType.php` also touched by #282 and #294.

Issue #285, opened 2026-09-11
[Bug]: The default design is created during a GET request and the exception catch cannot catch the unique constraint violation


loevgaard:
_Context: plugin 1.x @ 090e4ec._

### Severity
Medium (500 on every product page of the second channel without designs; DB writes on GET).

### What happened?
`GiftCardDesignProvider::getDesigns()` (`src/Provider/GiftCardDesignProvider.php:35-48`) is called from `GiftCardInformationType::buildForm()` (`src/Form/Type/GiftCardInformationType.php:44`), i.e. while rendering the product page on a **GET**. When the channel has no enabled design it creates a design with code `classic`, uploads an image and **flushes** (`:53-77`).

The `catch (ORMException $e)` at `:69` is dead code for the failure it describes: a duplicate code raises `Doctrine\DBAL\Exception\UniqueConstraintViolationException`, which extends `Doctrine\DBAL\Exception`, not `Doctrine\ORM\Exception\ORMException` (verified against doctrine/dbal 3.10 and doctrine/orm 2.20 in vendor). So:
- the **second channel** with zero designs hits the unique `code` (`GiftCardDesign.orm.xml`) → uncaught exception → 500 on every product page of that channel, with a closed EntityManager for the rest of the request;
- two concurrent first requests for the same channel do the same.
It also leaks the temp file created in `createTemporaryImageFile()` on failure.

### Suggested fix
- Stop creating data on GET. Ship the default design through the existing `setono_gift_card_design` fixture and/or a console command (`setono:gift-card:install-defaults`) documented in the README; make the design optional in the form (no designs → hide the picker; the card already renders the framed default when no image is present).
- If lazy creation is kept as a fallback: catch `\Doctrine\DBAL\Exception\UniqueConstraintViolationException` (and `\Throwable` around the upload), use a per-channel code (`classic_<channelcode>`), and re-query after the failure.

### Acceptance criteria
Functional test: two channels, no designs; rendering `GiftCardInformationType` for both does not throw. No DB writes happen while rendering the product page when designs exist.

### Overlaps for parallel work
`GiftCardInformationType.php` also touched by #282, #313, #314. Flush policy discussed in #300.

PR #329, opened 2026-09-14
Report a blank gift card amount or design position instead of crashing

### LINKED ISSUE #282: [Bug]: An empty amount (shop product page, admin gift card form) or empty design position throws a 500 instead of a validation error
_Context: plugin 1.x @ 090e4ec; Sylius 1.13 / 1.14, Symfony 6.4._

### Severity
High (shop checkout path breaks with a spinner and no message).

### What happened?
Three form fields map an empty submission (`null`) into a non-nullable setter. Symfony's `DataMapper` writes into the object through PropertyAccess during `submit()`, **before** the `POST_SUBMIT` validation listener, so PropertyAccess throws `InvalidArgumentException: Expected argument of type "int", "null" given` and the request ends in a 500 instead of a `NotBlank` violation:

1. Shop product page: `src/Form/Type/GiftCardInformationType.php:47-54` (`MoneyType`, empty → null) into `src/Order/GiftCardInformation.php:22` `setAmount(int $amount)`. The add-to-cart form posts via AJAX and Sylius' `sylius-add-to-cart.js` only understands the 400 shape (`response.errors.errors`), so the button keeps spinning with no feedback. Reproduced with a minimal form built from vendor classes; not covered by Playwright (only a valid `50` is ever submitted).
2. Admin gift card form: `src/Form/Type/GiftCardType.php:131-143` (`NumberType` + transformer returning `?int`) into `src/Model/GiftCard.php:146` `setAmount(int)`.
3. Admin design form: `src/Form/Type/GiftCardDesignType.php:31-34` (`IntegerType`, `required => false`) into `src/Model/GiftCardDesign.php:80` `setPosition(int)`.

### Suggested fix
- `GiftCardInformation`: make `$amount` nullable (`?int` property, `setAmount(?int)`, `getAmount(): ?int`) and let `NotBlank` report it; adjust `CartGiftCardHandler` (`Assert::notNull` before use) and `GiftCardInformationFactory`. Alternative: keep the model strict and bind the form to a nullable DTO.
- `GiftCardType`: `empty_data` or a transformer that never returns null into the setter; or `GiftCard::setAmount(?int)` with the XML `NotBlank` already present (`src/Resources/config/validation/GiftCard.xml:23-27`).
- `GiftCardDesignType`: `'empty_data' => '0'` for `position`, or nullable setter defaulting to 0.
- Add `TypeTestCase` tests submitting `''` for each field asserting a form error and no exception (see #309).
- Add a Playwright case submitting an empty amount on the product page and asserting the inline error.

### Acceptance criteria
Empty amount on the product page shows the "This value should not be blank" style error next to the field; empty amount/position in admin shows a field error; no 500s.

`GiftCardType.php` is also touched by #293 and #294; `GiftCardInformationType.php` by #285, #313, #314.

loevgaard:
## Summary

Three form fields mapped an empty submission into a non-nullable setter. Symfony's `DataMapper` writes submitted values through PropertyAccess during `submit()`, before the `POST_SUBMIT` validation listener runs, so `null` reaching `setAmount(int)`/`setPosition(int)` raised `InvalidArgumentException: Expected argument of type "int", "null" given` and the request ended in a 500 instead of a violation. On the shop product page that was entirely invisible to the customer: add to cart posts over AJAX and Sylius' `sylius-add-to-cart.js` only understands the 400 shape, so the button just kept spinning. The gift card information object now carries a nullable amount, so the `NotBlank` already declared on the field reports it; the two admin fields map an empty submission to `0`, which the constraints in `validation/GiftCard.xml` reject for a gift card and which is simply the default position for a design (the field is explicitly optional there). Constraints stay in XML, and nothing below the controllers changed.

Fixes #282

## Changes

- `src/Order/GiftCardInformation.php` / `GiftCardInformationInterface.php`: `$amount` is `?int` (property, constructor default, getter, setter), so a blank amount is representable until validation has run.
- `src/Cart/CartGiftCardHandler.php`: `Assert::notNull($amount)` before the amount is consumed — the handler only runs on a valid form.
- `src/Model/GiftCard.php` + interface and `src/Form/Type/GiftCardType.php`: `setAmount()` accepts `?int` (the property is null only while a blank submission is being validated; `getAmount()` still returns `int` and reports such a card as empty, so the money arithmetic is untouched), the form field no longer carries `empty_data`, and the existing `NotBlank` in `validation/GiftCard.xml` reports a blank amount. Changed after review from the initial `empty_data => 0`.
- `src/Model/GiftCardDesign.php` + interface and `src/Form/Type/GiftCardDesignType.php`: `position` is nullable on the model (setter and getter accept `?int`), the form field no longer carries `empty_data`, and a `NotNull` constraint in `validation/GiftCardDesign.xml` reports a blank position (message translated in en/da/fr). Changed after review from the initial `empty_data => 0`.
- `tests/Unit/Form/Type/`: new `TypeTestCase` tests for `GiftCardInformationType`, `GiftCardType` and `GiftCardDesignType`, each covering a valid submission plus the blank one (first form type tests in the repo, cf. #309).
- `tests/Playwright/specs/`: a shop case submitting an empty amount and asserting the inline message plus no 5xx, and two admin cases (issuing a card with no amount, saving a design with no position).

`GiftCardInformationFactory` needs no change: `OrderItemInterface::getUnitPrice()` already returns an `int`.

## Verification

All commands run from the worktree, against a dedicated database (`ssgc_blank_amount_*`) and port 8108.

- `vendor/bin/phpunit --testsuite unit` — OK (57 tests, 96 assertions)
- `vendor/bin/phpunit --testsuite functional` — OK (15 tests, 50 assertions)
- `composer analyse` — `[OK] No errors` (161 files, level max)
- `npx playwright test` (whole suite, `PLAYWRIGHT_BASE_URL=http://127.0.0.1:8108`) — 29 passed

Regression check — with `src/` stashed, on the unmodified code:

- the three new unit tests fail with exactly the reported exception: `Expected argument of type "int", "null" given at property path "amount"` (`GiftCardInformation::setAmount`, `GiftCard::setAmount`) and `... at property path "position"` (`GiftCardDesign::setPosition`);
- the three new Playwright cases fail (the shop one shows the empty validation banner the issue describes);
- against the served application, `POST /en_US/ajax/cart/add` with a blank amount returned **500**, `POST /admin/gift-cards/new` with a blank amount returned **500**, and `PUT /admin/gift-card-designs/1/edit` with a blank position returned **500**.

After the fix, the same three requests return **400** with `{"errors":{"errors":["This value should not be blank."]}}` (which is what `sylius-add-to-cart.js` renders), **422** with the field error *"The amount must be greater than or equal to 1"*, and **302** (saved with position `0`).

## Notes for reviewers

- `GiftCardInformationInterface::getAmount()` widened from `int` to `?int`. That is a signature change on a public interface, so a host application that implements it or type-hints the result needs the same widening. It is the option the issue recommends, and the alternative (a separate nullable DTO bound to the form) would have duplicated the model.
- The blank admin amount is reported by `GreaterThanOrEqual` ("The amount must be greater than or equal to 1") rather than by `NotBlank`, because `NotBlank` does not treat `0` as blank. The message is arguably better than "should not be blank" here, and no XML was touched.
- `src/Form/Type/GiftCardType.php` is also touched by in-flight #293 and #294, and `src/Form/Type/GiftCardInformationType.php` by #285/#313/#314 — this branch adds a single option line to `GiftCardType` and does not touch `GiftCardInformationType` at all, so conflicts should be trivial.
- The new `TypeTestCase` tests need a little scaffolding that #309 may want to factor out later: a Prophecy-backed `ManagerRegistry` for the design picker's `EntityType` (only the identifier metadata is read, the choices are passed explicitly), and a `ConstraintValidatorFactory` seeded with the validators that would otherwise need the container (`ValidGiftCardAmountValidator`, and a no-op for `UniqueEntity`, whose uniqueness check is a database lookup).


src/Form/Type/GiftCardDesignType.php
```diff
@@ -28,6 +28,9 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
             ->add('position', IntegerType::class, [
                 'label' => 'setono_sylius_gift_card.form.gift_card_design.position',
                 'required' => false,
+                // The field is optional, and the data mapper writes the submitted value into the non-nullable
+                // setter before validation runs, so a blank position means the default position, not null
+                'empty_data' => '0',
```
loevgaard: Isn't the 'Symfony way' of doing this to have a nullable setter and use validation instead?

loevgaard: Yes, that is the cleaner way. Done in a4e7343: `GiftCardDesign::$position` (and the interface) accept `?int`, the `empty_data` option is gone, and a `NotNull` constraint in `validation/GiftCardDesign.xml` reports a blank position with a translated message (en/da/fr). The unit test now asserts a form error and a null position instead of a silent 0, and the Playwright case asserts the validation error. For the gift card `amount` I left `empty_data` in place: `GiftCardInterface::getAmount(): int` feeds all the money arithmetic, so making it nullable would ripple through the whole plugin for one admin form; the XML `GreaterThanOrEqual(1)` already turns that blank into a field error. Happy to change that too if you prefer.

src/Form/Type/GiftCardType.php
```diff
@@ -69,6 +69,10 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
         });
         $builder->add('amount', NumberType::class, [
             'label' => 'sylius.ui.amount',
+            // The data mapper writes the submitted value into the non-nullable setter before validation runs,
+            // so a blank amount has to reach the model as a number the constraints in validation/GiftCard.xml
+            // reject rather than as null
+            'empty_data' => '0',
```
loevgaard: Maybe make this setter nullable as well?

loevgaard: Done in 2195578: `GiftCard::setAmount()` (and the interface) accept `?int`, `empty_data` is gone from the admin form, and the existing `NotBlank` constraint in `validation/GiftCard.xml` reports the blank. To keep the money arithmetic untouched, `getAmount()` still returns `int` and reports a card whose amount is null as empty; the property is only null between the form submit and validation. The unit test now asserts the NotBlank message on the field.

PR #330, opened 2026-09-14
Lock the currency of an issued gift card and validate it against the channel

### LINKED ISSUE #293: [Bug]: The currency of a live gift card can be changed in admin, and it is never validated against the channel

loevgaard:
## Summary

`GiftCardType` added `currencyCode` on every edit, while the channel and the amount are deliberately fixed once a card exists. That made the currency the one part of a live card's value an admin could still change: the balance and every `GiftCardTransaction` row are integers in the card's minor units and carry no currency of their own, so switching DKK to EUR on a redeemed card silently revalued it and the show page reformatted the whole ledger in the new currency. The currency is now only settable while the card is new; on edit the field is still rendered — the amounts on the page mean nothing without it — but `disabled`, so whatever is submitted for it is ignored.

Nothing validated that the currency belonged to the card's channel either, and the list offers every currency the shop knows (on a new card there is no channel yet, so the preferred-currency hint cannot help). An order is only ever priced in one of its channel's currencies, so a card issued outside them was rejected at the cart by `GiftCardIsApplicableValidator` with nothing having told the admin. A class-level `GiftCardCurrencyBelongsToChannel` constraint now reports that at the `currencyCode` field while the card can still be fixed, and a `Currency` constraint guards the code itself.

Fixes #293

## Changes

- `src/Form/Type/GiftCardType.php`: the `currencyCode` field is added `disabled` once the card has an id. The preferred-currency logic for new cards is untouched.
- `src/Validator/Constraints/GiftCardCurrencyBelongsToChannel.php` + `…Validator.php`: new class constraint reporting at path `currencyCode` when the card's currency is neither the channel's base currency (which the currency context falls back to) nor one of its listed currencies. Registered in `src/Resources/config/services/validator.xml`.
- `src/Resources/config/validation/GiftCard.xml`: the class constraint plus `NotBlank` and `Currency` on `currencyCode`, all in group `setono_sylius_gift_card` — the group only the admin gift card form uses, so nothing on the shop side changes.
- `src/Resources/translations/validators.{en,da,fr}.yml`: `gift_card.currency_code.{not_blank,invalid,not_in_channel}`.
- `tests/Unit/Validator/Constraints/GiftCardCurrencyBelongsToChannelValidatorTest.php`: new.
- `tests/Unit/Form/Type/GiftCardTypeTest.php`: new `TypeTestCase` — currency present and enabled on create (and submitting it works), present and disabled on edit (and a submitted value is ignored).
- `tests/Playwright/specs/admin/gift-cards.spec.js`: two specs — the currency select is enabled on `/new` and disabled on `/edit`, and issuing a card in a currency the channel does not offer comes back as a rendered, translated validation error. The second discovers the channel's currencies through a new `channelCurrencyCodes()` helper in `specs/support/fixtures.js` rather than assuming what the fixtures seeded.

## Verification

All run from the worktree, `vendor/bin` on PHP 8.4.

- `vendor/bin/phpunit --testsuite unit` — OK (58 tests, 90 assertions)
- `vendor/bin/phpunit --testsuite functional` — OK (15 tests, 50 assertions), against an isolated MySQL database
- `composer analyse` — [OK] No errors (PHPStan level max)
- `composer check-style` — [OK] No errors found
- `vendor/bin/rector process --dry-run` — [OK] Rector is done!
- `bin/console lint:yaml ../../src/Resources` — [OK] All 14 YAML files contain valid syntax
- `bin/console lint:twig ../../src/Resources` — [OK] All 18 Twig files contain valid syntax
- `bin/console lint:container` — [OK] The container was linted successfully
- `npx playwright test` against a served `tests/Application` on its own database — 28 passed (whole suite, admin and shop)
- Regression check on unmodified code (base files temporarily restored with `git checkout 1.x -- …`): `GiftCardTypeTest::it_shows_the_currency_but_locks_it_once_the_card_exists` fails, and both new Playwright specs fail — the second one actually creating a EUR card on a USD-only channel, which is the bug. Both pass again with the fix in place.

## Notes for reviewers

- `disabled` rather than dropping the field: an admin looking at a card needs to see what "1.000" on that page means, and a disabled field ignores submitted data, so it locks the value just as removing it would. The template already renders `form.currencyCode` unconditionally, so it needed no change.
- The class constraint accepts the channel's base currency even when it is not in the channel's `currencies` collection, because Sylius' currency context falls back to the base currency for an order. In the test app's fixtures the base currency is also listed, so this only matters for a channel configured the other way.
- The constraint validates on every admin submit, not only on create. On a card whose currency was legitimate when issued but whose channel has since dropped that currency, the admin will now have to put the currency back on the channel before they can edit the card. That seemed the right trade-off — such a card cannot be redeemed either — but it is the one behaviour change beyond the create path, so flagging it.
- Not implemented, as agreed: the issue's lower-priority point about `customer` being reassignable on a purchased card with no ledger trace. Worth a follow-up issue — either a zero-amount `manual` ledger row recording the reassignment, or an audit log entry.
- `src/Form/Type/GiftCardType.php` is also touched by in-flight #282 (the amount transformer) and #294. This branch only changes the currency listener, with no reformatting elsewhere in the file, so the hunks should not collide.


src/Validator/Constraints/GiftCardCurrencyBelongsToChannelValidator.php
```diff
@@ -0,0 +1,56 @@
...
+
+        if ($channel->getBaseCurrency()?->getCode() === $currencyCode) {
+            return;
+        }
+
+        foreach ($channel->getCurrencies() as $currency) {
+            if ($currency->getCode() === $currencyCode) {
+                return;
+            }
+        }
```
loevgaard: Can you explain why this is correct?

loevgaard: It was not correct, thanks for pushing on it. Sylius stores every order amount in the channel base currency; the other currencies a channel offers only change how amounts are displayed (the order carries currencyCode + exchangeRate for that). The plugin compares a card balance one to one with order totals, so the only currency a card can be issued in correctly is its channel base currency. Accepting any offered currency would let an admin issue a "EUR" card on a DKK channel whose balance would then be spent as DKK.

Changed in the latest commit: the constraint is now GiftCardCurrencyIsChannelBaseCurrency and only accepts the base currency (message key currency_code.not_base_currency in en/da/fr, unit test and the Playwright spec wording adjusted, everything green).

The same confusion exists elsewhere in code that predates this PR: CartGiftCardHandler and OrderGiftCardOperator::reconcile() stamp the card with the order display currency, and the applicability validator, applicator and coverage calculator compare the card currency with the order display currency instead of the channel base currency. I have opened a separate issue for that so this PR stays scoped to the admin form.

Issue #311, opened 2026-09-11
[Bug]: Cart gift card totals render above "Items total", outside the grid, and in the order currency rather than the display currency


loevgaard:
_Context: plugin 1.x @ 090e4ec._

### What happened?
- `src/Resources/views/shop/cart/_gift_card_totals.html.twig:6,10` emits `<div class="row">` blocks through the `sylius.shop.cart.summary.totals` event. In Sylius 1.13/1.14 that event fires **before** the totals `<table>` (vendor `SyliusShopBundle/Resources/views/Cart/Summary/_totals.html.twig:10-12`), so "Gift cards: -X / Remaining to pay: Y" appear above "Items total" rather than under "Order total", and the divs sit outside any `.ui.grid`, so Semantic UI does not lay them out as two columns.
- Amounts are formatted with `order.currencyCode` (`_gift_card_totals.html.twig:8,12`, `_gift_cards.html.twig:23`) while Sylius' totals use `money.convertAndFormat`, which converts to `sylius.currencyCode`; a shopper who switched display currency sees mixed currencies.
- The minus sign is a literal prefix rather than a negative amount through the formatter.
- The `sylius.shop.cart.summary` block uses priority 10, the same as Sylius' checkout button block, so its position relative to "Checkout" depends on registration order.

### Suggested fix
Render the two figures as rows of the totals table (a lower-priority block or a template override note), use the `money.convertAndFormat` macro, and give the summary block a distinct priority. Verify with a Playwright screenshot in `en_US` and `da_DK`.

### Overlaps
`redemption.spec.js` selectors (#308); cart templates also touched by #312, #287.

Issue #295, opened 2026-09-11
[Bug]: Email and PDF texts are translated in the request locale instead of the gift card's locale


loevgaard:
_Context: plugin 1.x @ 090e4ec._

### What happened?
Amounts in emails and PDFs are formatted with the passed `localeCode` (`src/Resources/views/email/gift_card.html.twig:12`, `src/Resources/views/shop/gift_card/pdf.html.twig:1`, resolved in `src/Pdf/DompdfGiftCardPdfGenerator.php:51-55`), but every `|trans` in `email/*.html.twig` (including the subjects at line 4), `pdf.html.twig` and the shared `shop/gift_card/_card.html.twig` omits the locale argument and therefore uses the translator's **current** locale: the admin's UI locale when an admin downloads or resends, the default locale from CLI/webhooks. A Danish customer's PDF downloaded by an English admin reads "GIFT CARD / VALUE / NO." around "kr. 500,00". Also:
- `pdf.html.twig:8` hardcodes `<html lang="en">`.
- `GiftCardEmailManager::sendGiftCard()` (`src/Mailer/GiftCardEmailManager.php:55`) uses the channel default locale, not the customer's.
- Expiry is printed as `|date('Y-m-d')` (`pdf.html.twig:70`), not localized.

Sylius' own email templates pass the locale explicitly (`|trans({}, null, localeCode)`).

### Suggested fix
- `|trans({}, 'messages', localeCode)` in email/PDF templates; for `_card.html.twig` pass `localeCode|default(app.request.locale)` from the product page include.
- `lang="{{ localeCode|replace({'_': '-'}) }}"`.
- Format the expiry with `format_date(pattern=..., locale=localeCode)` if `twig/intl-extra` is available (Sylius ships it), else document.
- Prefer the customer's locale in `sendGiftCard()` when the customer has one, falling back to the channel default.

### Acceptance criteria
Functional test: render the PDF for a card bought in `da_DK` while the translator locale is `en_US` and assert a Danish label (`messages.da.yml` has them) appears; same for the email subject.

### Overlaps
`pdf.html.twig` also touched by #302, #287; email templates by #303.

PR #332, opened 2026-09-14
Give the icon-only controls and the design picker accessible names

### LINKED ISSUE #312: [Bug]: Icon-only buttons and unlabeled inputs in the cart and admin grid have no accessible names

### What happened?
- `src/Resources/views/shop/cart/_gift_cards.html.twig:28-30`: the remove button is icon-only with no `aria-label`/`title`.
- `_gift_cards.html.twig:9` + `src/Form/Type/AddGiftCardToOrderType.php:31` (`'label' => false`): the code input has only a placeholder as its accessible name.
- `src/Resources/views/admin/gift_card/grid/field/customer.html.twig:5`: icon-only "open customer" link with no label and `target="_blank"` without `rel="noopener"`.
- `src/Resources/views/shop/product/show/_gift_card_information.html.twig:33-47`: `form_label(design)` is a `<label for>` pointing at a radio group (should be `fieldset`/`legend`); the picker has no `:checked` styling (`src/Resources/public/css/product-gift-card.css:8-16`), so the selected design is indicated only by the native radio dot; `_card.html.twig:43` emits `` in preview mode.

### Suggested fix
Visually-hidden labels / `aria-label`s, `fieldset`+`legend` for the picker with a visible selected state, `rel="noopener"`, omit the empty `img`. Add a Playwright accessibility check (axe) for the cart and product page.

### Overlaps
Same templates as #311, #313, #314.

loevgaard:
Stacked on #325 — it branches off `fix/design-provider`, whose `is defined and choices not empty` guard around the design picker block this change edits. Please merge #325 first; the base will then fold into `1.x`.

## Summary

Three controls the plugin adds to the storefront and the admin reached assistive technology with no name at all: the remove button in the cart and the "open customer" link in the gift card grid are icons with no text, and the redemption code field is rendered with `label => false` so a screen reader had only the placeholder to go on. They now carry translated `aria-label`s, and the customer link opts out of handing the tab it opens a `window.opener` handle.

The design picker on the product page was named by `form_label()`, a `<label for>` pointing at a radio group — a label can only ever point at one control, so the group had no accessible name and the label activated whichever radio came first. It becomes a `fieldset` with a `legend`, which is how a radio group is named, and the chosen design now has a visible state of its own instead of relying on the native radio dot alone. Finally, the live preview emitted ``, which resolves to the page URL and makes the browser fetch the whole document again as an image; the `src` is simply left off until the picker fills it in.

Fixes #312

## Changes

- `src/Resources/views/shop/cart/_gift_cards.html.twig`: a translated `aria-label` on the icon-only remove button (named after the card, so several applied cards stay apart) and an `aria-label` on the code field, merged into the field's existing `attr` so the placeholder survives. The form type keeps `label => false` — the inline `ui action input` layout has no room for a visible label, and an `aria-label` needs no stylesheet on the cart page.
- `src/Resources/views/admin/gift_card/grid/field/customer.html.twig`: `aria-label` + `title` on the icon-only customer link, and `rel="noopener"` on its `target="_blank"`.
- `src/Resources/views/shop/product/show/_gift_card_information.html.twig`: the picker is wrapped in `<fieldset><legend>`, the legend rendering whatever label the form type gives the field.
- `src/Resources/public/css/product-gift-card.css`: the fieldset is stripped back to the field it replaced and the legend styled like the Semantic UI labels around it, so the page looks unchanged; `:has(input:checked)` marks the selected design and `:focus-within` gives keyboard users an outline.
- `src/Resources/views/shop/gift_card/_card.html.twig`: the `src` attribute is only emitted when there is an image. The element itself stays, because the picker assigns its `src` live.
- `messages.{en,da,fr}.yml`: `ui.gift_card_code`, `ui.remove_gift_card`, `ui.open_customer`.
- `tests/Functional/CustomerGridFieldTest.php`: renders the admin grid field directly with and without a customer.
- `tests/Playwright/specs/shop/*.spec.js`: three new assertions, described below.

## Verification

Run from the worktree, on PHP 8.4, against a private database (`ssgc_a11y_*`) and port 8123.

- `vendor/bin/phpunit --testsuite unit` — OK (51 tests, 73 assertions)
- `vendor/bin/phpunit --testsuite functional` — OK (23 tests, 89 assertions)
- `bin/console lint:twig ../../src/Resources` — `[OK] All 18 Twig files contain valid syntax`
- `bin/console lint:container` — `[OK] The container was linted successfully`
- Playwright, against a freshly seeded `tests/Application` served with `symfony serve --no-tls --port=8123`:
  - `npx playwright test --project=shop` — 11 passed
  - `npx playwright test --project=admin` — 18 passed

Regression check, done by restoring the base templates with `git checkout fix/design-provider -- src/Resources` and running the new tests against them:

- `the design picker is a named group with a visible selection` — fails (no fieldset)
- `the live preview does not render an empty image source` — fails (``)
- `the redemption controls have accessible names` — fails (`getByLabel('Gift card code')` finds nothing)
- `CustomerGridFieldTest::it_names_the_icon_only_link_and_opens_it_safely` — fails (no `aria-label`)

All four pass again once the change is restored.

## Notes for reviewers

- **The admin link is covered by a functional test, not a spec.** The fixtures never assign a customer to a gift card, so the grid only ever renders the "No customer" branch in a browser and a Playwright assertion there would be dormant. `tests/Functional/CustomerGridFieldTest.php` renders the template through Twig with a persisted customer instead, which exercises the branch deterministically.
- **`img[src=""]` had to be asserted on the response, not the DOM.** The picker's JS assigns a `src` as soon as the page loads, so by the time a spec can look at the live DOM the empty `src` is gone — while the wasted request happens during parsing. The spec therefore fetches the HTML with `page.request.get()`.
- **The selected-design style uses `:has()`.** It is the only way to style the choice from the state of the radio inside it without asking the JS to mirror the selection into a class, and it is supported everywhere the rest of the plugin's CSS is. The spec asserts the computed `border-color` of the checked choice is not transparent rather than asserting a particular colour.
- **Files also touched by in-flight PRs.** `_gift_cards.html.twig` is edited by #311 (the amount cell) — this change only touches the widget call on line 9 and the button tag, so the hunks should not overlap. `_card.html.twig` is edited by #295/#314; the hunk here is the single `` line. `_gift_card_information.html.twig` sits on top of #325's guard, which is why this is stacked.
- Nothing here changes behaviour for sighted mouse users beyond the selected design now being outlined, and the legend renders in the same style as the `Amount` and `Message` labels around it.

## Added in review

- The "a design is required when the channel offers any" rule, which #325 left on the form field, moved into `validation/GiftCardInformation.xml` as the `GiftCardDesignRequired` constraint. Its validator asks `GiftCardDesignProviderInterface` what the channel offers, so a channel without designs still does not require one. Unit test for the validator; the functional tests from #325 for both cases pass unchanged.
- Rebased onto the current `1.x` (the branch used to carry #325's commits). After the rebase: unit OK (139 tests), functional OK (41 tests), PHPStan, ECS, Rector and the lints `[OK]`, Playwright 48 passed on fresh fixtures.


src/Form/Type/GiftCardInformationType.php
```diff
@@ -70,7 +77,8 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
 
                     return ['data-image-path' => $front?->getPath() ?? ''];
                 },
-                'constraints' => [
+                // A channel without designs skips the picker, so a design is only mandatory when there is one to pick
+                'constraints' => [] === $designs ? [] : [
                     new NotBlank(['groups' => self::VALIDATION_GROUPS]),
                 ],
```
loevgaard: Doesn't constraints belong in XML files?

loevgaard: They do. That rule came in with #325 (this branch was stacked on it): it sat on the field because a design is only mandatory when the channel offers any, which a plain `NotBlank` in the mapping cannot express. It is now a `GiftCardDesignRequired` constraint in `validation/GiftCardInformation.xml`, whose validator asks the design provider what the channel offers, and the form type declares no rules of its own any more (only the `Valid` cascade into its data object, which is wiring rather than a rule). The functional tests from #325 for both cases pass unchanged through the mapping, and there is a unit test for the validator. Also rebased onto the current `1.x`, so the diff is down to this PR's own changes.

Issue #305, opened 2026-09-11
[Bug]: Admin-created gift cards are emailed even when disabled or expired, and there is no resend action


loevgaard:
_Context: plugin 1.x @ 090e4ec._

### What happened?
- `SendGiftCardEmailSubscriber::onGiftCardPostCreate()` (`src/EventSubscriber/SendGiftCardEmailSubscriber.php:36-40`) emails the card whenever a customer is set and `sendNotificationEmail` is true, which defaults to true (`src/Model/GiftCard.php:55`). It sends even when the admin created the card **disabled** or with an expiry in the past.
- There is no way to resend: no "Send email" action on the show page or grid, although docblocks mention "admin resend" (`GiftCardEmailManagerInterface.php:21`, `DompdfGiftCardPdfGenerator.php:47`). A customer who lost the email cannot re-download either (the PDF route is admin-only, `routes/admin.yaml:49-58`).

### Suggested fix
- Skip sending when `!$giftCard->isUsable()` (or at least `!isEnabled()`), and say so in the form help text.
- Add an admin "Send email" action (POST + CSRF) on the show page and grid that calls `GiftCardEmailManager::sendGiftCard()` and flashes the result.

### Overlaps
`show.html.twig`/grid also touched by #315; email content by #303.

Issue #315, opened 2026-09-11
[Feature request]: Admin gift card list and show page: status, filters, originating order, applied orders, delete visibility, redirect after adjustment


loevgaard:
### Describe the proposed solution
Grid (`src/DependencyInjection/SetonoSyliusGiftCardExtension.php:169-324`):
- Filters are only `code` and `enabled` (`:230-239`). Add channel, customer (email contains), currency, delivery type, expired (`expiresAt < now`), spent (`amount = 0`), created-at range, and a toggle to reveal pending cards (hidden by `GiftCardRepository::createListQueryBuilder()`, `src/Doctrine/ORM/GiftCardRepository.php:23`, with no way back).
- `customer` and `amount` are not sortable (`:189-203`).
- Replace the `enabled` column with a status column (usable / expired / spent / pending / disabled).
- The delete action is shown for non-deletable cards and only blocked after the click by `GiftCardDeletionSubscriber` (`:36`); use a custom action template that checks `isDeletable()`.
- Bulk actions are enabled on the route (`routes/admin.yaml:6`) but none are defined.
- Design grid: `name` not sortable, no channels column.

Show page (`src/Resources/views/admin/gift_card/show.html.twig`):
- Link the originating order (`GiftCard::getOrder()` exists at `src/Model/GiftCard.php:118`, never rendered) and list `appliedOrders` (`:201`).
- Status badge; design thumbnail/link; "Send email" (#305).
- `AdjustGiftCardBalanceAction.php:61` redirects to the update page instead of the show page where the ledger is; `adjust_balance.html.twig:18` breadcrumb likewise.
- Customer column (`grid/field/customer.html.twig`) links the email to `mailto:` and the customer page via an icon; the show page does the reverse (`:58`). Make them consistent.

### Overlaps
`show.html.twig` also touched by #301, #305; `prepend()` grid config by #281.

PR #333, opened 2026-09-14
Do not email unusable gift cards, and add an admin send action

loevgaard:
Its parent #324 has been merged into 1.x, so this PR now targets 1.x directly.

## Summary

An admin creating a gift card got an email sent to the customer whenever a customer was set and the notification checkbox was ticked — which it is by default — no matter what state the card was in. A card created disabled, or with an expiry already in the past, was emailed anyway, so the customer received a gift that does not work. There was also no way back: nothing in the admin could send the card once it became usable, or resend one the customer lost, even though the docblocks on `GiftCardEmailManagerInterface` and `DompdfGiftCardPdfGenerator` already spoke of admin resends.

The post-create subscriber now skips a card that is not usable (`isUsable()`: disabled, expired or empty), and the notification checkbox says when nothing is sent. Sending is instead offered explicitly: a "Send email" action on the gift cards grid and on the show page, next to "Download PDF". Sending reaches the customer, so — like the scaffold action from #324 — it is a POST carrying a CSRF token rather than a link, which a browser prefetch or an `` on any page the admin visits would otherwise follow. A card with no customer email address is reported as such rather than silently doing nothing, because `GiftCardEmailManager::sendGiftCard()` returns without sending in that case.

Fixes #305

## Changes

- `SendGiftCardEmailSubscriber` returns without sending when `!$giftCard->isUsable()`.
- `GiftCardType` gives the `sendNotificationEmail` checkbox a `help` text saying the email only goes out for a usable card (`send_notification_email_help`, en/da/fr).
- New `Controller/Action/Admin/SendGiftCardEmailAction`: validates the CSRF token (403 otherwise), 404s on an unknown card, calls `GiftCardEmailManagerInterface::sendGiftCard()`, flashes success or, without a customer email address, an error, and redirects to the show page.
- Route `setono_sylius_gift_card_admin_gift_card_send_email` (`POST /gift-cards/{id}/send-email`, `_sylius.section: admin`, `permission: true`) and the service in `services/controller.xml`.
- Grid item action `send_email` of type `setono_sylius_gift_card_post_link` (prepended grid config only), asking for confirmation first, and a matching POST form button on `admin/gift_card/show.html.twig`.
- Flash keys `gift_card.email_sent` / `gift_card.email_not_sent_no_customer` in `flashes.{en,da,fr}.yml`, label `ui.send_email` in `messages.{en,da,fr}.yml`.
- Tests: `tests/Unit/EventSubscriber/SendGiftCardEmailSubscriberTest` (usable sends; disabled, expired, empty, unticked and customer-less do not), `tests/Unit/Controller/Action/Admin/SendGiftCardEmailActionTest` (forged token → 403 and nothing sent, no repository hit; valid token → sent, success flash, redirect to the show page; no customer email → nothing sent, error flash), and four new Playwright specs.

## Verification

- `vendor/bin/phpunit --testsuite unit` — OK (62 tests, 98 assertions).
- `vendor/bin/phpunit --testsuite functional` — OK (15 tests, 50 assertions), on its own database.
- `composer analyse` — `[OK] No errors` (PHPStan max).
- `composer check-style` — `[OK] No errors found`.
- `vendor/bin/rector process --dry-run` — `[OK] Rector is done!`.
- `bin/console lint:yaml ../../src/Resources` — all 14 files valid; `lint:twig ../../src/Resources` — all 19 files valid; `lint:container` — linted successfully.
- Playwright, whole suite against a served `tests/Application` on its own port and database (`MAILER_DSN=null://null`) — **31 passed**, including the four new specs: the show page carries a `send-email` POST form with a CSRF token and submitting it lands back on the show page with a flash; the grid offers sending as a form and not as an `a[href$="/send-email"]`; a GET returns 405 and a POST with a forged token 403; the create form explains when the notification is not sent.
- Regression check on the subscriber: with `src/EventSubscriber/SendGiftCardEmailSubscriber.php` restored from the base branch, `SendGiftCardEmailSubscriberTest` fails 3 of 6 (disabled, expired, no balance) and passes with the fix.
- The success path of the action was additionally exercised by hand against the served app — a seeded card temporarily given a customer produced the "The gift card was emailed to the customer" flash, with the PDF generated and handed to the (null) mailer.

## Notes for reviewers

- The seeded gift cards have no customer, so the Playwright spec reads the customer off the show page and asserts whichever flash matches, rather than assuming what the fixtures produced.
- The action deliberately does not refuse an unusable card: sending is the admin's explicit decision, and refusing it would remove the only way to hand over a card whose state the admin has just fixed. Only the automatic send at creation is gated on `isUsable()`.
- The gift cards grid now carries six item actions, and the actions column already overflowed its cell with five (measured at 1500px viewport: 1594px before this change, 1720px after; both fit at 1920px). Worth a follow-up — possibly folded into #315, which also touches this grid.
- `src/Resources/views/admin/gift_card/show.html.twig` and the grid config are also touched by #315, and `src/Form/Type/GiftCardType.php` by #282/#293 — only the `sendNotificationEmail` checkbox definition is changed here.
- The customer-facing half of #305 (letting a customer re-download their own gift card PDF; the PDF route is admin-only) is not addressed here — it needs a shop-side route with its own authorization and is a separate change.


loevgaard:
GitHub closed this when its parent branch was deleted after #324 merged. Reopened as #336 against 1.x with the same branch and description.

PR #334, opened 2026-09-14
Send the customer back to the cart when a gift card is redeemed concurrently

### LINKED ISSUE #288: [Bug]: Concurrent redemption of one gift card surfaces as a 500 (OptimisticLockException) at checkout completion

### What happened?
`GiftCard` has a version column (`src/Resources/config/doctrine/model/GiftCard.orm.xml:18`, `version="true"`), which is the right tool: when two orders redeem the same card at the same moment, the second flush's `UPDATE ... WHERE version = ?` affects 0 rows and Doctrine throws `OptimisticLockException`. Nothing in the plugin handles it (`grep -ri "OptimisticLock\|LockMode" src` is empty), so the customer gets a 500 page from the checkout "complete" request. Their cart still has the card applied, and because the other order drained it, a retry runs into #278.

### Suggested fix
Pick one:
- **Catch and redirect**: a kernel exception subscriber (or a decorator around the checkout complete flow) that catches `Doctrine\ORM\OptimisticLockException` during the shop checkout complete request, re-processes the cart, flashes "One of your gift cards changed while you were checking out, please review your cart" and redirects to `sylius_shop_cart_summary`. Cheap, and it composes with the guard from #278.
- **Pessimistic lock**: in `PaymentRedemptionMethod::commit()` re-load each gift card with `LockMode::PESSIMISTIC_WRITE` inside an explicit transaction around the transition. More invasive because the resource controller owns the flush.

### Acceptance criteria
A functional test that bumps the card's `version` via DQL between apply and complete asserts: no 500, order not placed, flash present, cart intact. The unit `GiftCardBalanceOperatorTest` unchanged.

#278 (same flow). Test scaffold: #306.

loevgaard:
## Summary

`GiftCard` carries a version column, so two orders redeeming the same card at the same moment race on it: the loser's `UPDATE ... WHERE version = ?` matches no row and Doctrine raises an `OptimisticLockException` out of the flush that completes the checkout. The issue expected a 500 there. On a stock Sylius shop it is not one: `Sylius\Bundle\CoreBundle\Doctrine\ORM\Handler\ResourceUpdateHandler` wraps the exception in a `RaceConditionException`, which `ResourceController::updateAction` handles by redirecting to the referer. Nothing was written (the transaction rolled back, the order is not placed, the cart still holds the card), so the race is already harmless.

An earlier revision decorated `sylius.resource_controller.resource_update_handler` to let the Doctrine exception through at checkout completion so the plugin could add a message. Per review that override is gone: it put the plugin in the path of every resource update in the application and traded a redirect Sylius guarantees for one that depends on our subscriber. Sylius' handling is left exactly as it is, and this PR now pins it with a test and adds a safety net for applications where the exception does reach the kernel.

Fixes #288

## Changes

- `src/EventSubscriber/GiftCardRaceConditionSubscriber.php`: a `kernel.exception` subscriber for `sylius_shop_checkout_complete` only, and only when an `OptimisticLockException` is anywhere in the exception chain. It adds the `setono_sylius_gift_card.gift_card.changed_during_checkout` error flash and redirects to `sylius_shop_cart_summary`. It runs at priority `-64`, between Symfony's `logKernelException` (0) and the error page rendering (-128), so the race is still logged. On a stock shop it never fires; it exists for applications that complete checkout with a controller of their own or replaced the update handler. The entity manager is closed once a flush fails, so it does no database work.
- Registered in a new `services/event_subscriber.xml` (imported from `services.xml`), per the rule from the #338 review. #338 adds the same file, so whichever merges second gets a trivial conflict.
- New flash message in `flashes.en.yml`, `flashes.da.yml` and `flashes.fr.yml`.
- Tests: `tests/Unit/EventSubscriber/GiftCardRaceConditionSubscriberTest.php`; `tests/Functional/GiftCardRedemptionRaceTest.php` bumps `version` and drains the balance with a DQL `UPDATE` (writing past the identity map exactly like a competing request) and pins two things: the flush really raises `OptimisticLockException` and the losing redemption writes no ledger row, and at checkout completion the real update handler from the container turns it into a `RaceConditionException` whose previous exception is the Doctrine one.

## Verification

From the worktree, rebased onto the current `1.x`, own database.

- `vendor/bin/phpunit --testsuite unit` — OK (145 tests, 477 assertions)
- `vendor/bin/phpunit --testsuite functional` — OK (43 tests, 215 assertions)
- `composer analyse`, `composer check-style`, `vendor/bin/rector process --dry-run` — all `[OK]`
- `bin/console lint:container` — `[OK]`; `debug:container sylius.resource_controller.resource_update_handler` resolves to `Sylius\Bundle\CoreBundle\Doctrine\ORM\Handler\ResourceUpdateHandler`, i.e. nothing of ours decorates it

## Notes for reviewers

- **Where the customer-facing message comes from.** `sylius_shop_checkout_complete` sets `flash: false`, so Sylius' own race condition flash is not shown and the loser is redirected back to the complete step without a word. With #338 in, that redirect runs into its check on the complete step: a card the winner drained is removed with a message and the customer is sent to the cart; a card that now covers less gets the gateway payment re-sized, also with a message; and if the winner left enough balance the customer simply places the order again. No override needed for any of it.
- The subscriber walks the whole `getPrevious()` chain rather than matching only the top exception, so it also fires if some other layer wraps the Doctrine exception.


src/Doctrine/ORM/Handler/CheckoutRaceConditionUpdateHandler.php
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * Sylius wraps the OptimisticLockException raised by the flush in a RaceConditionException, which the resource
+ * controller swallows: it adds an error flash and redirects to the referer. The shop's checkout complete route
+ * sets `flash: false`, so no flash is added either, and the customer silently lands back on the checkout page
+ * with no idea why the order was not placed.
+ *
+ * Only that one route is affected — everywhere else, and for any other update handling failure, Sylius' own
+ * behaviour is left untouched.
+ */
+final class CheckoutRaceConditionUpdateHandler implements ResourceUpdateHandlerInterface
```
loevgaard: We are changing some fundamentals with this override, aren't we?

loevgaard: We were, and it was not worth it, so it is gone. The decorator sat in the path of every resource update in the application (scoped to one route in behaviour, but global in reach), relied on Sylius' internal wrapping and decoration order, and traded a redirect Sylius guarantees for one that only happens if our subscriber is there to catch the exception we let loose.

What is left: Sylius' handling is untouched (`debug:container sylius.resource_controller.resource_update_handler` resolves to Sylius' own class again). A lost race at checkout completion becomes a `RaceConditionException`, the resource controller redirects to the referer, and the functional test now pins exactly that instead of expecting the Doctrine exception to escape. The `kernel.exception` subscriber stays as a safety net for applications where the Doctrine exception does reach the kernel (own checkout controller, replaced update handler); on a stock shop it never fires, and its docblock says so. It is registered in `services/event_subscriber.xml`, per your rule from #338.

The message the decorator was after comes for free once #338 is in: the redirect lands on the complete step again, where #338's check finds the card drained or covering less, tells the customer and sends them to the cart. If the winner left enough balance, the customer simply places the order again.

src/EventSubscriber/GiftCardRaceConditionSubscriber.php
```diff
@@ -0,0 +1,113 @@
...
+
+    public function onKernelException(ExceptionEvent $event): void
+    {
+        $request = $event->getRequest();
+
+        if (self::CHECKOUT_COMPLETE_ROUTE !== $request->attributes->get('_route')) {
+            return;
+        }
+
+        if (!self::causedByOptimisticLock($event->getThrowable())) {
```
loevgaard: This fires for any optimistic lock failure on the complete route, not only a gift card's. Sylius locks the order number sequence (`SequentialOrderNumberGenerator`) and product stock (`OrderInventoryOperator`) optimistically as well. So in an application where the exception reaches the kernel, two customers checking out at the same moment get "One of your gift cards changed" even when neither has a gift card.

`OptimisticLockException::getEntity()` names the entity that lost, so checking that it is a `GiftCardInterface` keeps this to gift card races.

src/EventSubscriber/GiftCardRaceConditionSubscriber.php
```diff
@@ -0,0 +1,113 @@
...
+            if ($session instanceof Session) {
+                $session->getFlashBag()->add('error', self::FLASH);
+            }
+        }
+
+        $event->setResponse(new RedirectResponse($this->urlGenerator->generate(self::CART_SUMMARY_ROUTE)));
+
+        // Symfony's own error listener overwrites whatever response was set before it, so the redirect only
+        // survives if the event stops here
+        $event->stopPropagation();
```
loevgaard: Nit: `setResponse()` already stops propagation (`RequestEvent::setResponse()` calls `stopPropagation()`), so this line is redundant and the reason given in the comment above does not apply.

tests/Functional/GiftCardRedemptionRaceTest.php
```diff
@@ -0,0 +1,209 @@
...
+    private function updateHandler(): ResourceUpdateHandlerInterface
+    {
+        /** @var ResourceUpdateHandlerInterface $handler */
+        $handler = self::getContainer()->get(ResourceUpdateHandlerInterface::class);
+
+        return $handler;
+    }
+
+    /**
+     * The request has to be the current one, because that is how the handler knows which route it is serving
```
loevgaard: Nit: left over from the decorator. Sylius' own update handler does not read the request stack, so this no longer holds, and the pushed request stays on the stack for the rest of the test.

loevgaard (commented):
Reviewed at e3760ec. Leaving Sylius' update handler alone was the right call. One real issue and two nits inline, plus:

- **Merge after #338.** On a stock shop this subscriber never runs, so the message the customer gets for #288 comes from #338's check on the complete step.
- **Worth a thought, not verified:** two orders redeeming the same card at the same moment can also end in an InnoDB deadlock rather than a version mismatch. Doctrine runs inserts before updates, so each order's ledger insert takes a shared lock on the gift card row through the foreign key, and each then needs an exclusive lock for the versioned update. That surfaces as a DBAL `DeadlockException`, which neither Sylius' handler nor this subscriber catches.

src/EventSubscriber/GiftCardRaceConditionSubscriber.php
loevgaard: Right, and in exactly the case this is for, an application whose checkout lets the Doctrine exception through, both customers would have been told about gift cards they don't have. Fixed in 00561b7: `lostGiftCardRace()` (was `causedByOptimisticLock()`) still walks the chain, but at the first `OptimisticLockException` it returns `$exception->getEntity() instanceof GiftCardInterface`. A race lost on the order sequence or product stock is left to Symfony's error handling as before.

New unit test `it_ignores_a_race_lost_on_anything_but_a_gift_card` uses `OptimisticLockException::lockFailed(new OrderSequence())`, the exception the flush raises, and asserts no response, no stopped propagation and no flash. The existing positive tests now pass a `GiftCard` as the entity instead of a `stdClass`.

src/EventSubscriber/GiftCardRaceConditionSubscriber.php
loevgaard: Removed in 707dace, comment included. The unit test still asserts `isPropagationStopped()` after the redirect, which now just pins what `setResponse()` does: Symfony's error listener at -128 never gets to replace the redirect.

loevgaard:
Addressed the review at e3760ec. The three inline points are answered in their threads. This covers the review body.

**Merge after #338.** Added at the top of the description. On a stock shop the subscriber never runs. The customer-facing message for #288 comes from #338's check on the complete step, which is where Sylius' `RaceConditionException` redirect sends the customer.

**Deadlock: it is real, and this PR now fixes it (40d54a3).**

These are the statements the checkout-completion flush issues, captured from Sylius' real update handler with the checkout `complete` transition:

```
INSERT INTO sylius_payment (...)
INSERT INTO setono_sylius_gift_card__gift_card_transaction (..., gift_card_id, order_id, payment_id)   -- FK check: S lock on the card's row
UPDATE setono_sylius_gift_card__gift_card SET amount = ?, updated_at = ?, version = version + 1 WHERE id = ? AND version = ?   -- needs X
```

That is the order you described, and I reproduced the deadlock in two ways:

- **One real flush against a competing connection.** The competing connection held the S lock from its own ledger row, then sent its versioned update while the real flush sat between its inserts and its updates. In `SHOW ENGINE INNODB STATUS`, both transactions wait for `lock_mode X locks rec but not gap` on the card's PRIMARY record, and each conflicts with the other's `lock mode S locks rec but not gap`. The checkout got a `Doctrine\DBAL\Exception\DeadlockException` on that `UPDATE`, with no `OptimisticLockException` anywhere in the chain.
- **Two real checkouts in two processes.** Both used Sylius' update handler with the `complete` transition, on the same card, with the gap between inserts and updates widened to 0.7 s. Order numbers were preset, so the fresh test database's missing order sequence couldn't collide first. One of the two deadlocked in 3 of 3 runs on MySQL 8.0.46 and in 3 of 3 runs on MariaDB 11.6.2.

I also found a second problem along the way. MariaDB turned on `innodb_snapshot_isolation` by default in 11.6.2. There, even the plain sequential race doesn't raise an `OptimisticLockException`. In that race the other order commits after this one has started reading but before it writes. The loser's ledger insert then fails with `1020 Record has changed since last read`, which is a plain `DriverException`. So on MariaDB 11.6+, #288 was an error page on a stock shop even without the deadlock.

**What changed:**

- `PaymentRedemptionMethod::commit()` now takes a `PESSIMISTIC_WRITE` lock on each card it is about to redeem, before anything of the order is written.
  - The locks are taken in id order, so two orders sharing several cards can't deadlock on the locks themselves.
  - They are taken inside the transaction that Sylius' update handler already wraps the transition and the flush in.
- The second order waits for the first to commit. Its versioned update then matches no row, which raises the `OptimisticLockException` that Sylius turns into a redirect.
- Some databases report the loss at the lock instead. Those failures are thrown as `OptimisticLockException::lockFailed($giftCard)`, so they end up in the same redirect and the subscriber recognises them too:
  - 1020 on MariaDB 11.6+.
  - A deadlock with a writer that did not lock first (an admin adjustment, a cancel/refund restore).
- Any other database error passes through untouched.
- Outside a transaction no lock is taken, since nothing would hold it until the flush.
- d887b1d adds `doctrine/dbal` to `require`, with the same constraint `doctrine/orm` uses, because the plugin now names DBAL classes and the dependency analysis job flagged it as a shadow dependency.

After the change I ran the same two-process experiment 3 times on each database. Every time, one order was placed and the other got a `RaceConditionException` wrapping an `OptimisticLockException`.

**New tests:**

- `it_holds_the_gift_card_before_writing_anything_at_checkout_completion` (functional): when this order's ledger row is about to be written, a second connection with a 1 s lock wait timeout cannot insert its own ledger row for the card. Without the lock it fails on both databases.
- `it_reaches_sylius_as_a_race_condition_when_the_gift_card_changes_during_checkout_completion` (functional): another connection redeems the card and commits after the order has started reading inside Sylius' transaction. The test expects a `RaceConditionException` whose previous exception is an `OptimisticLockException` naming the card.
  - On MariaDB 11.6 it fails both without the lock (1020 from the ledger insert) and without the conversion (1020 from the lock).
  - On MySQL it passes either way, because MySQL reports that case at the version check.
- `PaymentRedemptionMethodTest` (unit):
  - Locks are taken in id order before anything is redeemed.
  - No lock is taken outside a transaction.
  - A deadlock or a 1020 at the lock becomes an `OptimisticLockException` naming the card.
  - A lock wait timeout passes through as is.

**Not covered:** the admin balance adjustment and the cancel/refund restore still write their ledger row before they update the card. Two of those racing each other can still deadlock. That happens on the admin side, not at checkout. The admin adjustment flushes outside any transaction, so it can't take the lock the same way, and today it doesn't handle a lost version check either. I kept that out of this PR. Say if you want an issue for it.

**Rebase.** Rebased onto `1.x` at 73eaa78, which includes #352, with no conflicts.

- The subscriber only implements Symfony's `EventSubscriberInterface` and takes `router`, so it needs no interface alias.
- The new constructor argument on `PaymentRedemptionMethod` is `doctrine`, which is not a plugin service.
- `InterfaceAliasWiringTest` passes.

**Verification after the rebase and the review changes:**

- `vendor/bin/phpunit --testsuite unit`: OK (154 tests, 618 assertions)
- `vendor/bin/phpunit --testsuite functional`: OK (46 tests, 226 assertions) on MariaDB 11.6.2, and OK (46 tests, 226 assertions) on MySQL 8.0.46, the version CI uses
- `composer analyse`, `composer check-style`, `vendor/bin/rector process --dry-run`: all `[OK]`
- `bin/console lint:container` in dev and test: `[OK]`
- I didn't run Playwright: nothing rendered changed, and no spec completes a checkout.

loevgaard:
Re-reviewed: the entity check, the redundant `stopPropagation()`, the stale test comment and the deadlock are all addressed, and the locking looks right to me.

Rebased onto 1.x after #338, #342 and #340. Three conflicts, resolved as follows:
- `services/event_subscriber.xml` (added here and by #338): both services kept.
- `PaymentRedemptionMethod`: git merged `commit()` itself (lock before the payments, #338's `assertPaidInFull()` after them); the conflict was only the private methods, and #342's `rollbackPayment()`/`assertPaidInFull()` and this PR's `lockGiftCards()`/`lostToAnotherWriter()` are all kept.
- `tests/Unit/Redemption/PaymentRedemptionMethodTest.php` (added here and by #342): one file with this PR's five locking tests and #342's four refund tests. The refund tests build the method through their own `rollbackRedemptionMethod()`, which passes the new `ManagerRegistry` argument.

Locally on the rebased branch: unit (227 tests), functional (71 tests, MariaDB 11.6), PHPStan, ECS and `composer validate` all pass. Merging once CI is green.

Issue #314, opened 2026-09-11
[Bug]: The 500-character message limit overflows the card, and line breaks are dropped


loevgaard:
_Context: plugin 1.x @ 090e4ec._

### What happened?
`src/Form/Type/GiftCardInformationType.php:59-64` allows 500 characters (`maxlength` + `Length`), with no visible hint or counter. The card's hero block is bottom-anchored (`src/Resources/views/shop/gift_card/_card_style.html.twig:23`), so a long message pushes the 52px title into the brand row in both the live preview and the PDF. Newlines are collapsed: the message is rendered in a `<div>` without `nl2br`/`white-space: pre-line` (`_card.html.twig:28,46`) and the preview writes `textContent`. Admin `GiftCardType` has no limit at all.

### Suggested fix
Pick a limit the card can hold (~200) or make it configurable (`purchase.max_message_length`) and apply it in both forms; show a counter; `white-space: pre-line` in preview and PDF; clamp/step down the font in the PDF for long messages.

### Overlaps
Same files as #313, #312; `GiftCardInformationType.php` by #282.

PR #337, opened 2026-09-14
Show the gift card amount limits and read the typed amount in the page locale

### LINKED ISSUE #313: [Bug]: Product page preview parses decimal commas as integers, and the amount field shows no limits and no numeric keypad

### What happened?
- `src/Resources/public/js/product-gift-card.js:32` does `parseFloat(amountInput.value)`, so `50,50` typed in da/fr becomes 50 and the live preview shows 50,00; `Intl.NumberFormat` with `data-locale` can also differ from `sylius_format_money`.
- `:17` and `_gift_card_information.html.twig:6` fall back to `USD` when the channel has no base currency; `:20` binds to the first `input[type=text]` in the container (fragile).
- The amount field shows no minimum/maximum although `GiftCardAmountLimitsProvider` exists; the customer discovers limits from the post-submit error (`ValidGiftCardAmountValidator.php:53-67`). No `inputmode="decimal"`, so no numeric keypad on mobile. Preset chips are #272.

### Suggested fix
- Show `help` with the formatted min/max (reuse the money formatter as the validator does) and set `attr: {inputmode: 'decimal', 'data-js-gc-amount-input': ''}`; bind the JS to that attribute.
- Parse using the locale's decimal separator (`Intl.NumberFormat(locale).formatToParts(1.1)`), or read the value Sylius' money widget produces.
- Cover with a Playwright case in `da_DK`.

### Overlaps
Same template/JS as #312, #314; `GiftCardInformationType.php` also by #282, #285.

loevgaard:
## Summary

The gift card amount field on the product page had three problems. The live preview ran the typed value through `parseFloat`, so `50,50` in a comma decimal locale previewed as `50,00`, and it bound to the first text input it found in the container rather than to the amount field itself. The field advertised neither the minimum nor the maximum the validator enforces, so the customer only discovered the limits from the error after submitting. And it asked for no numeric keypad on mobile.

The amount field now carries a translated help text quoting the limits from `GiftCardAmountLimitsProvider`, formatted with Sylius' money formatter the way the validator formats them, an `inputmode="decimal"` and a `data-js-gc-amount-input` hook. The preview binds to that hook and parses the typed value with the decimal separator the page locale uses, so what the customer types is what the card shows. The money in the help text is formatted in the locale being browsed, the way the shop formats its prices everywhere else.

Fixes #313

## Changes

- `src/Form/Type/GiftCardInformationType.php`: inject `GiftCardAmountLimitsProviderInterface`, `MoneyFormatterInterface` and `LocaleContextInterface`; the `amount` field gets `help` (minimum, or the range when a maximum is configured) plus `attr` `inputmode="decimal"` and `data-js-gc-amount-input`. The help is omitted when the channel has no base currency, since there would be nothing to format the limits as.
- `src/Resources/config/services/form.xml`: the three new arguments.
- `src/Resources/translations/messages.{en,da,fr}.yml`: `amount_help_minimum` and `amount_help_range`.
- `src/Resources/public/js/product-gift-card.js`: bind to `[data-js-gc-amount-input]` (falling back to the old selector for host apps that render their own field), read the locale's decimal separator from `Intl.NumberFormat(locale).formatToParts(1.1)` and parse with it, and keep working without `Intl` (a dot separator and `toFixed(2)`).
- `src/Resources/views/shop/product/show/_gift_card_information.html.twig`: an absent channel base currency no longer falls back to `USD`; the empty currency makes the preview format a plain number instead of pretending the shop sells in dollars.
- `tests/Playwright/specs/shop/gift-card-product.spec.js`: two cases — the field asks for a numeric keypad and shows help quoting a money figure, and typing `50,50` in a comma decimal locale previews `50,50`. Locale discovery moved into a helper that also reads the `switch-locale/<locale>` links, which is how the switcher actually exposes the other locales; the existing "renders in every locale" case now uses it too and therefore really visits them.
- `tests/Application/config/packages/setono_sylius_gift_card.yaml`: the fixture channel only carried the default locale, so no comma decimal locale existed to test against. It now offers `en_US` and `fr_FR`.

## Verification

- `vendor/bin/phpunit --testsuite unit` — OK (51 tests, 73 assertions)
- `vendor/bin/phpunit --testsuite functional` — OK (15 tests, 50 assertions), own database `ssgc_js_preview_test`
- `composer analyse` — [OK] No errors
- `composer check-style` — [OK] No errors found
- `vendor/bin/rector process --dry-run` — [OK] Rector is done!
- `bin/console lint:yaml ../../src/Resources` / `lint:yaml config` / `lint:twig ../../src/Resources` / `lint:container` — all OK
- `npx playwright test` against a served test app (own database and port) — 28 passed, the whole suite including admin
- Regression check: with `src/` restored to `1.x` (`git checkout 1.x -- src/...`) the two new cases fail (`inputmode` missing, preview shows `50,00`), and pass again with the change in place
- Manually confirmed both help branches render: `Choose any amount from $1.00.` / `Choisissez un montant d'au moins 1,00 $US.`, and with a maximum configured `Choose any amount between $1.00 and $2,500.00.` / `Choisissez un montant compris entre 1,00 $US et 2 500,00 $US.`

## Notes for reviewers

- `ValidGiftCardAmountValidator` still formats its messages without a locale, so the validation error quotes `$1.00` where the help quotes `1,00 $US` on a French page. Aligning the validator is a separate change and out of scope here.
- The parser treats only the locale's decimal separator as a decimal point and drops every other non-digit, so `1.250,50` in a comma locale previews as 1250,50. A dot typed in a comma locale is read as a grouping separator, matching what the server does with the value.
- `GiftCardInformationType.php` is also touched by #282 and #285 — this change only adds constructor arguments and the `amount` field options. `_gift_card_information.html.twig` and `product-gift-card.js` are also touched by #312 and #314; only the container/amount lines are changed here, the design picker block is untouched.
- The test application now seeds a second channel locale. It makes the existing "renders in every locale" case meaningful as well, and the full Playwright suite passes with it.


tests/Application/config/packages/setono_sylius_gift_card.yaml
```diff
@@ -5,6 +5,16 @@ sylius_fixtures:
...
             fixtures:
+                channel:
+                    options:
+                        custom:
+                            fashion_web_store:
+                                # Sylius seeds the channel with the default locale only. The shop specs need a
+                                # locale that writes decimals with a comma to cover the live gift card preview
+                                locales:
+                                    - en_US
+                                    - fr_FR
```
loevgaard: This makes `sylius:fixtures:load` fail at random with `Column 'name' cannot be null` in the `order` fixture, which is the End to end failure on this PR. I reproduced it locally: two of three loads failed.

The failing statement inserts an empty `en_US` row into `sylius_product_translation` for a product that already has one. The order fixture picks the order locale from the channel's locales, and `ProductRepository::findLatestByChannel()` fetch-joins only that locale's translation. A product loaded for a French order therefore holds only its French translation. When something later asks it for `en_US`, `TranslatableTrait::getTranslation()` finds neither that nor the fallback, creates an empty English translation, and the flush fails on its null name.

So the comma-decimal case needs its locale from somewhere other than the channel the demo orders are generated in.

loevgaard (commented):
Reviewed at 2cd5c40. The form type, the translations and the parser look right. Two things block it:

- **Fixture loading fails at random** because of the new channel locale, which is what the End to end job hits. Details inline.
- **It conflicts with 1.x.** #339 moved the amount constraints out of `GiftCardInformationType` into `validation/GiftCardInformation.xml` and added a constructor argument. When rebasing, keep the XML constraints and do not bring the form-level `constraints` back.

tests/Application/config/packages/setono_sylius_gift_card.yaml
loevgaard: Reproduced on the rebased branch before changing anything: 2 of 5 loads failed with `Column 'name' cannot be null`. I traced the insert to confirm the path: the product held only its `fr_FR` translation (loaded by `findLatestByChannel()` for a French order and kept in the identity map), and `OrderItemNamesSetter` asked it for `en_US` while completing a later English order, so `getTranslation()` created the empty English row.

Fixed in 9ab677d by taking the second locale off the channel again. The test app seeds Sylius' default single-locale channel and the order fixture only ever generates `en_US` orders. The comma decimal case now takes its locale from the page instead: the preview reads it from the `data-locale` attribute the template writes on `#setono-gift-card-information`, so the spec loads the real product page and rewrites that attribute to `fr-FR` through `page.route()` on its way to the browser. The markup, the field and the script are the shop's own. Before that, on the unmodified page, it checks that the template writes the browsed locale (`en-US`) and that a dot is read as the decimal separator there. The case no longer skips when the shop offers no comma locale, so it always runs.

What I ruled out:

- Rendering a locale the channel does not offer: `NonChannelLocaleListener` redirects the shop to the homepage.
- A second channel with the comma locale and the order fixture pinned to `FASHION_WEB`: the suite reaches the shop at 127.0.0.1, which matches no channel hostname and only resolves because `SingleChannelContext` applies while there is one channel. A second channel means giving `FASHION_WEB` a hostname and the new one another that the spec cannot discover from the shop, and the product fixtures would put every demo product and its prices into the new channel too.
- Adding the locale in a fixture that runs after `order`: it is still the channel the orders are generated in, only later, and it breaks again as soon as something after it loads products per locale.

The server side of the locale is covered by the form type unit test, which asserts the help quotes the minimum as the money formatter returns it for the locale context's `fr_FR`.

After the change 8 of 8 loads in a row succeeded (plus the two loads for the Playwright runs), and the full Playwright suite passes on fresh fixtures (50 passed). With `product-gift-card.js` put back to 1.x's, the new case fails with `Received string: "50,00 $US"`.

loevgaard:
Rebased onto `1.x` (on top of #352). The PR commit is now 364217b and the fixture fix is 9ab677d; the latter is answered in the inline thread.

- #339's constraints stay in `validation/GiftCardInformation.xml`. The `amount` field has no form-level `constraints` again; this PR only adds its `attr` (`inputmode="decimal"`, `data-js-gc-amount-input`) and the help options to 1.x's field. The three new constructor arguments come after #339's `$maximumMessageLength`, so the existing argument order is unchanged.
- Per #352, `form.xml` injects the limits provider as `GiftCardAmountLimitsProviderInterface`, not the class id. The other two new arguments are Sylius' `sylius.money_formatter` and `sylius.context.locale`. `InterfaceAliasWiringTest` passes.
- #339 added a unit test that constructs `GiftCardInformationType` directly, so it now passes the new collaborators, and it has one more case asserting the help (`amount_help_minimum` with the minimum as the formatter returns it for the locale context's `fr_FR`), `inputmode` and the preview hook.
- The translations, the template and the script merged cleanly next to #339's message counter and #332's design fieldset. Nothing from this PR was dropped apart from the form-level constraints and the fixture channel's `fr_FR` locale.

Unit, functional, PHPStan, ECS, Rector, `lint:container`/`lint:twig`/`lint:yaml` and the full Playwright suite are green after the rebase; the numbers are in the description under "Added in review".

loevgaard:
Re-reviewed: the fixture and rebase points are addressed. After rebasing onto 1.x, codecov/patch fell to 68.00% against a 68.22% baseline because only the minimum-only help branch was tested. I added 8ae59fc, which builds the form again with a maximum configured and with a channel without a base currency, so `GiftCardInformationType` is fully covered. Merged once all 26 checks were green.

Issue #284, opened 2026-09-11
[Bug]: Refunds are not handled: refunding a gift card payment does not restore the balance and refunding a purchase order does not disable the cards


loevgaard:
_Context: plugin 1.x @ 090e4ec; Sylius 1.13 / 1.14._

### Severity
High (money inconsistency; documented behaviour does not exist).

### What happened?
`README.md:41` and `CLAUDE.md` state that balances are "restored when it is cancelled/refunded". Only `sylius_order.cancel` is hooked (`SetonoSyliusGiftCardExtension::prepend()` lines 91-111 and `src/EventSubscriber/Workflow/RollbackRedemptionSubscriber.php`). `grep -rni refund src` matches nothing except a docblock in `GiftCardBalanceOperatorInterface`.

Two gaps:
1. **Refunding a gift-card payment does not restore the balance.** An admin refunds the payment whose method is the gift card method (Sylius admin "Refund" → `sylius_payment.refund`), or the whole order payment state moves to `refunded`. The payment becomes `refunded`, the customer's card stays drained, and the ledger records nothing.
2. **Refunding a gift-card purchase order does not disable the cards.** `OrderGiftCardOperator::disable()` runs only on `sylius_order.cancel`. After `sylius_order_payment.refund` on an order that bought gift cards, the merchant has returned the money and the cards remain enabled and spendable.

### Suggested fix
1. Hook `sylius_payment` `refund` (after) on both adapters: if `GiftCardPaymentChecker::isGiftCardPayment($payment)`, restore via `GiftCardBalanceOperator::restore($card, $payment->getAmount(), $order, $payment, $key)`. Extract the per-payment part of `PaymentRedemptionMethod::rollback()` into a `rollbackPayment(PaymentInterface)` used by both paths, and align the idempotency key (`rollback()` uses `restore:order:%s:payment:%s` and itself applies `refund` to the payment, so the new hook must be a no-op for that case).
2. Hook `sylius_order_payment` `refund` (after) → `OrderGiftCardOperator::disable($order)` for orders that bought gift cards. Decide explicitly what `partially_refund` does (probably nothing) and document it.
3. Register both in `prepend()` (winzou) and as subscribers in `src/EventSubscriber/Workflow/`; keep the two in sync as CLAUDE.md requires.
4. Fix the README/CLAUDE.md wording to describe what is actually restored and when.

### Acceptance criteria
[...]

PR #338, opened 2026-09-14
Refuse to complete a checkout whose applied gift cards no longer cover it

### LINKED ISSUE #278: [Bug]: Order is placed as paid with zero payments when an applied gift card becomes unusable before checkout completes
_Context: plugin 1.x @ 090e4ec; Sylius 1.14.19 (vendor). Applies to 1.13 as well._

### Severity
**Critical – customers can receive goods for free.** Proven by a functional test run against `tests/Application` and MySQL (test below).

### What happened?
1. A customer applies gift card `G` (balance 5000) to a cart that totals 5000. The checkout order-processor decorator removes the gateway payment because the card covers everything (`src/OrderProcessor/GiftCardAwareOrderPaymentProcessor.php:47`), and the requirement checker lets the checkout skip the payment step (`src/Checker/GiftCardAwarePaymentMethodSelectionRequirementChecker.php:26`).
2. Before the customer presses "Place order", `G` becomes unusable: it is spent from another cart (another device, or another person holding the same code), an admin disables/expires it, or its balance is adjusted.
3. The `complete` transition runs. Sylius does **not** re-run the order processor on `complete` (`sylius_process_cart` is only registered for address/select_shipping/select_payment/skip_* in vendor `sylius_order_checkout.yml`). `PaymentRedemptionMethod::commit()` (`src/Redemption/PaymentRedemptionMethod.php:59`) recomputes coverage from the card's **live** balance (`src/Calculator/GiftCardCoverageCalculator.php:31-34`), finds 0 and creates no payment.
4. Sylius' `OrderPaymentStateResolver::getTargetTransition()` returns `pay` when `$order->getPayments()->isEmpty()`. The order is placed as **paid** (and **fulfilled**, since nothing is shippable) with no payment of any kind.

Why the ordering is what it is: winzou sorts callbacks ascending by priority. Sylius' checkout callbacks run at -400 (`sylius_create_order`, which cascades `sylius_order.create` and therefore the plugin's `commit`) and -200 (`sylius_control_payment_state`); the plugin's own callbacks have priority 0 and run last. The Symfony Workflow subscribers mirror this.

A genuine race makes this worse: two carts committing the same card simultaneously → the second flush throws `OptimisticLockException` (see #288), the customer retries, and the retry lands exactly here. An attacker needs one card worth V and N carts to get (N-1)·V of goods for free.

### Reproduction (functional test, extends `GiftCardFunctionalTestCase`)
```php
<?php
declare(strict_types=1);
namespace Setono\SyliusGiftCardPlugin\Tests\Functional;

use Setono\SyliusGiftCardPlugin\Factory\GiftCardFactoryInterface;
use Setono\SyliusGiftCardPlugin\Model\GiftCardInterface;
use Setono\SyliusGiftCardPlugin\Tests\Application\Model\{Order, OrderItem, OrderItemUnit, Product};
use Sylius\Abstraction\StateMachine\StateMachineInterface;
use Sylius\Component\Core\Model\{Customer, PaymentInterface, ProductVariant};
use Sylius\Component\Core\{OrderCheckoutStates, OrderCheckoutTransitions, OrderPaymentStates};

final class StaleGiftCardAtCheckoutCompletionTest extends GiftCardFunctionalTestCase
{
    /** @test */
    public function a_usable_gift_card_pays_the_order_at_checkout_completion(): void
    {
        $giftCard = $this->createEnabledGiftCard('CONTROL000000001', 5000);
        $order = $this->createCheckoutReadyOrder(5000);
        $order->addGiftCard($giftCard);
        $this->manager->flush();

        $this->stateMachine()->apply($order, OrderCheckoutTransitions::GRAPH, OrderCheckoutTransitions::TRANSITION_COMPLETE);
        $this->manager->flush();

        self::assertSame(OrderPaymentStates::STATE_PAID, $order->getPaymentState());
        self::assertCount(1, $order->getPayments());
        self::assertSame(PaymentInterface::STATE_COMPLETED, $order->getPayments()->first()->getState());
        self::assertSame(0, $giftCard->getAmount());
    }

    /** @test */
    public function an_order_whose_applied_gift_card_became_unusable_must_not_be_placed_as_paid(): void
    {
        $giftCard = $this->createEnabledGiftCard('STALE00000000001', 5000);
        $order = $this->createCheckoutReadyOrder(5000);
        $order->addGiftCard($giftCard);
        $this->manager->flush();

        $giftCard->setAmount(0); // spent elsewhere / adjusted / (or ->disable(), or expiresAt in the past)
        $this->manager->flush();

        $this->stateMachine()->apply($order, OrderCheckoutTransitions::GRAPH, OrderCheckoutTransitions::TRANSITION_COMPLETE);
        $this->manager->flush();

        // On 1.x today these three assertions FAIL in the opposite direction: checkoutState=completed,
        // payments=0, paymentState=paid, state=fulfilled.
        self::assertNotSame(OrderPaymentStates::STATE_PAID, $order->getPaymentState());
        self::assertTrue($order->getPayments()->count() > 0 || OrderCheckoutStates::STATE_COMPLETED !== $order->getCheckoutState());
    }

    private function stateMachine(): StateMachineInterface
    {
        return self::getContainer()->get('sylius_abstraction.state_machine');
    }

    private function createEnabledGiftCard(string $code, int $amount): GiftCardInterface
    {
        /** @var GiftCardFactoryInterface $factory */
        $factory = self::getContainer()->get('setono_sylius_gift_card.factory.gift_card');
        $giftCard = $factory->createForChannel($this->getChannel());
        $giftCard->setCode($code);
        $giftCard->setInitialAmount($amount);
        $giftCard->setAmount($amount);
        $giftCard->enable();
        $this->manager->persist($giftCard);

        return $giftCard;
    }

    /** An ordinary product in a cart that skipped the payment step, i.e. where a fully covered cart sits */
    private function createCheckoutReadyOrder(int $unitPrice): Order
    {
        $product = new Product();
        $product->setCurrentLocale('en_US'); $product->setFallbackLocale('en_US');
        $product->setCode('MUG'); $product->setName('Mug'); $product->setSlug('mug');
        $this->manager->persist($product);

        $variant = new ProductVariant();
        $variant->setCurrentLocale('en_US'); $variant->setFallbackLocale('en_US');
        $variant->setCode('MUG_VARIANT'); $variant->setName('Mug');
        $variant->setProduct($product); $variant->setShippingRequired(false);
        $this->manager->persist($variant);

        $customer = new Customer();
        $customer->setEmail('review@example.com');
        $this->manager->persist($customer);

        $order = new Order();
        $order->setChannel($this->getChannel()); $order->setCurrencyCode('USD'); $order->setLocaleCode('en_US');
        $order->setCustomer($customer);
        $order->setCheckoutState(OrderCheckoutStates::STATE_PAYMENT_SKIPPED);

        $item = new OrderItem();
        $item->setVariant($variant); $item->setUnitPrice($unitPrice);
        new OrderItemUnit($item);
        $order->addItem($item);
        $this->manager->persist($order);

        return $order;
    }
}
```
Observed on 1.x: `checkoutState=completed`, `payments=0`, `paymentState=paid`, `state=fulfilled`, `total=5000`.

### Root cause
Coverage is derived from the card's live balance at every step and nothing is reserved or re-validated between "apply" and "complete". The `setono_sylius_gift_card__order_gift_cards` join table stores no amount, and `commit()` silently skips cards whose coverage dropped instead of failing.

### Suggested fix
- Add a **guard** on `sylius_order_checkout.complete`: a winzou `guard` callback in `SetonoSyliusGiftCardExtension::prepend()` and a `workflow.sylius_order_checkout.guard.complete` subscriber in `src/EventSubscriber/Workflow/`. The guard re-runs the applicability rules used by `GiftCardIsApplicableValidator` for every card in `$order->getGiftCards()` (usable, same channel, same currency) and recomputes coverage. If any card is no longer usable or the coverage changed compared to what the order's payments assume, remove that card, re-process the order (`sylius.order_processing.order_processor`), block the transition, and add a flash so the customer lands on the cart with an explanation.
- Make `commit()` **throw** when `coverage total < order total − sum(non-gift-card payments)` instead of continuing. A placed order must never carry less payment than its total.
- Longer term (shared with #279): persist the applied amount per (order, gift card) or hold balance while a card is applied to a cart.

### Acceptance criteria
- The test above fails on 1.x and passes after the fix; the control test keeps passing.
- Both adapters covered (winzou via `sylius_abstraction.state_machine`, Symfony Workflow via `state_machine.sylius_order_checkout`).
- Cancelling/expiring/adjusting an applied card through the admin UI and then completing the cart shows a flash and returns to the cart.

Touches `src/Redemption/PaymentRedemptionMethod.php`, `src/DependencyInjection/SetonoSyliusGiftCardExtension.php` (`prepend()`), `src/Resources/config/services/workflow.xml`. Also touched by #279, #284, #296, #288. Test scaffold shared with #306.

loevgaard:
## Summary

Nothing re-validated a gift card between the cart and "Place order". A cart the card fully covered skipped the payment step and carried no gateway payment, so a card spent from another cart, disabled or adjusted in the meantime left `commit()` with nothing to redeem, no payment was created, and Sylius' payment state resolver, finding no payments at all, placed the order as paid — goods for free (the functional test from the issue reproduces it). The checkout's `complete` transition is now guarded on both state machine adapters: a winzou `guard` callback in `prepend()` and a `workflow.sylius_order_checkout.guard.complete` subscriber ask `GiftCardCoverageGuard` whether every applied card can still pay for the order and whether, together with the order's other payments, they add up to its total. The eligibility rules the validator and the applicator each carried are extracted into `GiftCardEligibilityChecker`, so the guard shares one source of truth with them, and `commit()` throws `UnderpaidOrderException` instead of quietly placing an underpaid order should anything reach it another way. For the customer, a `kernel.request` subscriber corrects a stale cart on the complete step (removes the cards that can no longer be used, or re-sizes the gateway payment when they all still can but cover less), flashes why and sends them back to the cart to start checkout over, where the payment step is no longer skipped. It builds on the explicit callback priorities merged in #319 and extends that PR's parity test to guards.

Fixes #278

## Changes

- `src/Checker/GiftCardEligibilityChecker` (+ interface, `GiftCardIneligibilityReason` enum): the enabled / expired / balance / channel / currency rules, now used by `GiftCardIsApplicableValidator`, `GiftCardApplicator` and the guard.
- `src/StateMachine/GiftCardCoverageGuard` (+ interface): `isSatisfiedBy()` for the state machine, `getIneligibleGiftCards()` and `isTotalCovered()` for the correction. Orders without gift cards pass; that is Sylius' business.
- `prepend()`: `callbacks.guard.setono_sylius_gift_card__guard_gift_card_coverage` on `complete` (priority 0, with the reason stated: Sylius registers no guard on `complete`). `EventSubscriber/Workflow/GuardCheckoutCompletionSubscriber` mirrors it and blocks the `GuardEvent`.
- `EventSubscriber/CheckoutGiftCardCoverageSubscriber` on `kernel.request` (priority 4) for requests whose `_sylius.state_machine` is `sylius_order_checkout` / `complete`: corrects the cart, flushes, flashes (`setono_sylius_gift_card.gift_card.no_longer_usable` with `%code%` grouped in fours, the way the cart lists the code, or `coverage_changed`) and redirects to `sylius_shop_cart_summary`.
- `PaymentRedemptionMethod::commit()`: after creating the gift card payments, throws `UnderpaidOrderException` when the payments on an order with gift cards add up to less than its total. Idempotent: on a repeat call the existing gift card payments count instead of the balances they consumed.
- Flash translations in en, da, fr.
- Tests: `tests/Functional/StaleGiftCardAtCheckoutCompletionTest.php` (the issue's control and hole tests on winzou, the same through `state_machine.sylius_order_checkout`, a reduced-balance variant, and `commit()` throwing); unit tests for the checker, the guard, the request subscriber and the workflow guard subscriber; `StateMachineCallbackParityTest` now knows the `guard` position, builds the right workflow event per kind and stubs collaborator return values so a `bool`-returning guard can be dummied.
- Playwright: `tests/Playwright/specs/shop/stale-gift-card.spec.js` issues a card through the admin, walks a guest to the complete step, disables the card in the admin and asserts both "Place order" and a plain reload of the complete page land on the cart with the flash and without the card. The shop project now depends on the `setup` project so it can reuse the stored admin session.

## Verification

- `vendor/bin/phpunit --testsuite unit` — OK (76 tests, 373 assertions)
- `vendor/bin/phpunit --testsuite functional` (own database `ssgc_stale_coverage_test`) — OK (23 tests, 118 assertions)
- `vendor/bin/phpunit tests/Functional/StaleGiftCardAtCheckoutCompletionTest.php` — OK (6 tests, 42 assertions); with the `src` changes stashed the four hole tests fail (`Failed asserting that true is false` / no `UnderpaidOrderException`) and the two control tests pass
- `composer analyse` — OK, no errors
- `composer check-style` — OK
- `vendor/bin/rector process --dry-run` — OK
- `bin/console lint:yaml ../../src/Resources` (14 files), `lint:twig` (18 files), `lint:container` — all OK
- `bin/console debug:config winzou_state_machine sylius_order_checkout` shows the guard; `debug:event-dispatcher kernel.request` shows the subscriber at priority 4 ahead of `CheckoutResolver` at 0
- Playwright, whole suite against `symfony serve --port=8119` on a freshly seeded dev database — 29 passed (1.5m), including the two new journeys
- All of the above re-run after rebasing onto the current `1.x` (#317, #319, #324 merged)

## Notes for reviewers

- **Why `kernel.request` and not the `sylius.order.pre_complete` resource event or a validation constraint.** Sylius' `CheckoutResolver` (a `kernel.request` subscriber at priority 0) asks the state machine `can()` for the route's transition on every request to a checkout page and, told no, redirects to the route mapped to the cart's checkout state — for `payment_skipped` that is the complete step itself. With a guard in place, a plain GET of `/checkout/complete` with a stale cart would redirect forever. So the correction has to run before the resolver, which the resource event (dispatched in the controller) and a class constraint on the Order in the `sylius_checkout_complete` group (validated in the controller) cannot do; the request-level subscriber covers GET and PUT alike, and the second Playwright journey pins the no-loop behaviour. The guard stays the hard stop for anything that reaches the transition another way.
- The window between the request-level check and the transition is the same request; a card spent in exactly that window is caught by the guard and surfaces as a `StateMachineExecutionException`. Sylius' own `OrderIntegrityChecker` on `pre_complete` re-processes the order first, so in that case the gateway payment is re-sized and the order is not placed underpaid anyway.
- The guard and the `commit()` check only apply to orders that have gift cards applied. Vanilla Sylius places a zero-total order without payments as paid; that is left alone.
- The validator's message precedence changed in one corner: a card that is already applied *and* mismatches channel or currency now reports the mismatch rather than "already applied".
- Also touched by in-flight work: `SetonoSyliusGiftCardExtension::prepend()` (#284 will add refund callbacks; the new `guard` section sits above `after` in `sylius_order_checkout`), `PaymentRedemptionMethod.php`, `StateMachineCallbackParityTest.php` (extended, not restructured), `services/redemption.xml`, `services/workflow.xml`, `services/event_subscriber.xml` (new). `services/redemption/payment.xml` and `GiftCardAwareOrderPaymentProcessor` (#320) are untouched.
- Longer term, as the issue says, persisting the applied amount per (order, gift card) or holding balance while a card is applied would remove the need to recompute at completion; out of scope here.


src/Checker/GiftCardApplicabilityChecker.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Setono\SyliusGiftCardPlugin\Checker;
+
+use Setono\SyliusGiftCardPlugin\Model\GiftCardInterface;
+use Sylius\Component\Core\Model\OrderInterface;
+
+final class GiftCardApplicabilityChecker implements GiftCardApplicabilityCheckerInterface
```
loevgaard: Usually Sylius uses the terminology of 'eligible' instead of 'applicable'. Wdyt?

src/StateMachine/GiftCardCoverageGuard.php
```diff
@@ -0,0 +1,63 @@
...
+namespace Setono\SyliusGiftCardPlugin\Guard;
+
+use Setono\SyliusGiftCardPlugin\Calculator\GiftCardCoverageCalculatorInterface;
+use Setono\SyliusGiftCardPlugin\Checker\GiftCardApplicabilityCheckerInterface;
+use Setono\SyliusGiftCardPlugin\Model\OrderInterface;
+use Setono\SyliusGiftCardPlugin\Payment\GiftCardPaymentCheckerInterface;
+use Sylius\Component\Core\Model\OrderInterface as CoreOrderInterface;
+use Sylius\Component\Core\Model\PaymentInterface;
+
+final class GiftCardCoverageGuard implements GiftCardCoverageGuardInterface
```
loevgaard, line 14: Does this belong in a `StateMachine` folder maybe?

src/Resources/config/services/event_listener.xml
```diff
@@ -30,5 +30,17 @@
                  class="Setono\SyliusGiftCardPlugin\EventListener\PendingGiftCardCleanupListener">
             <tag name="doctrine.event_listener" event="onFlush"/>
         </service>
+
+        <service id="Setono\SyliusGiftCardPlugin\EventSubscriber\CheckoutGiftCardCoverageSubscriber"
```
loevgaard: Subscribers should be in `event_subscriber.xml`. If there are others, please create an issue about it

src/Checker/GiftCardApplicabilityChecker.php
loevgaard: Agreed, renamed. The checker is now `Checker\GiftCardEligibilityChecker` (+ `GiftCardEligibilityCheckerInterface::getIneligibilityReason()`) and the enum `GiftCardIneligibilityReason`; the guard's method is `getIneligibleGiftCards()`. The pre-existing `GiftCardIsApplicable` / `BalanceAdjustmentIsApplicable` constraints on 1.x are untouched here, happy to rename those the same way in a follow-up if you want.

src/StateMachine/GiftCardCoverageGuard.php
loevgaard: Yes, it exists for the checkout state machine's `complete` transition. Moved to `Setono\SyliusGiftCardPlugin\StateMachine\GiftCardCoverageGuard` (+ interface), test under `tests/Unit/StateMachine`.

src/Resources/config/services/event_listener.xml
loevgaard: Moved into a new `services/event_subscriber.xml` (imported from `services.xml`). There are others: two more in `event_listener.xml`, one each in `mailer.xml` and `menu.xml`, and the workflow ones in `workflow.xml`; listed in #346.

src/Checker/GiftCardApplicabilityChecker.php
loevgaard: Please rename those as well

loevgaard (commented):
Reviewed at adbd6c3. The guard, the request subscriber and the `commit()` check hold up against the redemption flow, and CI is green.

- **Still open from the earlier review:** renaming `GiftCardIsApplicable` and `BalanceAdjustmentIsApplicable` to the eligibility wording. "Please rename those as well" came after the last push.
- One leftover copy of the eligibility rules, inline.
- Nit: `stale-gift-card.spec.js` copies `addSomethingToCart()` and `applyGiftCard()` from `redemption.spec.js`. They could move to `specs/support`.

src/Checker/GiftCardEligibilityCheckerInterface.php
```diff
@@ -0,0 +1,20 @@
...
+
+use Setono\SyliusGiftCardPlugin\Model\GiftCardInterface;
+use Sylius\Component\Core\Model\OrderInterface;
+
+interface GiftCardEligibilityCheckerInterface
+{
+    /**
+     * Why the gift card cannot pay for the order, or null when it can. The same rules decide whether a card may be
+     * applied to a cart and whether a card already applied may stay on it as checkout completes, so the validator,
+     * the applicator and the checkout guard all ask here rather than each keeping a copy of them.
```
loevgaard: `GiftCardCoverageCalculator::isEligible()` still keeps its own copy of these rules (`isUsable()` plus the currency, without the channel check), and `commit()` sizes the gift card payments from it. The guard stops a channel mismatch first, so nothing is wrong today, but the calculator could ask this checker as well and make this sentence true.

src/Checker/GiftCardApplicabilityChecker.php
loevgaard: Renamed in 53de299: `GiftCardIsApplicable` → `GiftCardIsEligible` and `BalanceAdjustmentIsApplicable` → `BalanceAdjustmentIsEligible`, with their validators (`GiftCardIsEligibleValidator`, `BalanceAdjustmentIsEligibleValidator`, which are also the service ids) and the two validation mappings (`AddGiftCardToOrderCommand.xml`, `AdjustGiftCardBalanceCommand.xml`). The message keys never carried the "applicable" wording, so they are unchanged. Both constraints were added by the 1.x rewrite, so there is no released name to note in UPGRADE-1.0.md. `pplicab` no longer appears in src/, tests/ or the docs, apart from "Conditions générales applicables" in the French card terms.

`BalanceAdjustmentIsEligible` only checks that an adjustment does not take the balance below zero. If you'd rather the name said that (e.g. `BalanceAdjustmentKeepsBalanceNonNegative`), I'll change it.

src/Checker/GiftCardEligibilityCheckerInterface.php
loevgaard: Done in 4bf6a57. `GiftCardCoverageCalculator` now takes `GiftCardEligibilityCheckerInterface` as a second constructor argument, wired through the interface alias. It counts a card only when `getIneligibilityReason($giftCard, $order)` returns null, and `isEligible()` is gone. So a card from another channel now covers nothing anywhere coverage is used (`commit()`, the gateway payment sizing in `GiftCardAwareOrderPaymentProcessor`, the payment step skip), not only behind the guard. The checker has no dependencies, so the new argument creates no cycle.

This docblock now lists the calculator among the callers. `GiftCardCoverageCalculatorInterface`'s docblock points at the checker instead of listing the rules itself. `GiftCardCoverageCalculatorTest` stubs the checker with Prophecy: one case per `GiftCardIneligibilityReason`, plus one showing that a rejected card leaves its share to the next card.

loevgaard:
Addressed the review at adbd6c3 and rebased onto `1.x` (now at 73eaa78, including #352). New head: 699d1af.

- **Rename:** `GiftCardIsApplicable` → `GiftCardIsEligible` and `BalanceAdjustmentIsApplicable` → `BalanceAdjustmentIsEligible`, validators and service ids included (53de299; details in the thread).
- **Leftover copy of the eligibility rules:** the coverage calculator now asks the checker (4bf6a57; details inline).
- **Nit:** `addSomethingToCart()` and `applyGiftCard()` moved to `tests/Playwright/specs/support/cart.js`, with the gift card field selector, and both `redemption.spec.js` and `stale-gift-card.spec.js` use them (699d1af). The shared `applyGiftCard()` is the redemption spec's version: it only submits the code, because the redemption spec also submits codes that must be rejected. The stale card journeys now assert on their own that the cart lists the card.
- **Rebase:** #352's wiring rules are applied in the commits that introduce the services, not in a fix-up. The winzou guard callback now calls `@Setono\SyliusGiftCardPlugin\StateMachine\GiftCardCoverageGuardInterface`. That alias is public and the `GiftCardCoverageGuard` class id no longer is, the same way `OrderGiftCardOperatorInterface` is wired. Every new argument names an interface alias: the eligibility checker in the applicator, validator and guard; the coverage calculator and payment checker in the guard; the guard, applicator and code normalizer in `CheckoutGiftCardCoverageSubscriber`; the guard in `GuardCheckoutCompletionSubscriber`. `StateMachineCallbackParityTest` maps the guard's collaborator to the interface alias, and it passes alongside `InterfaceAliasWiringTest`. The priorities are unchanged (guard 0 on both adapters).

Verification after the rebase and the changes above: unit (167 tests), functional (48 tests), `composer analyse`, `composer check-style`, `rector --dry-run`, `lint:container` (dev and test), `lint:yaml`, `lint:twig` all pass. The whole Playwright suite passed 50/50 against a freshly seeded database.

loevgaard:
One more commit, found while restacking #340 on this branch: fe02a98 makes `stale-gift-card.spec.js` stop assuming the US can be shipped to.

The journeys picked the US as the shipping country on the assumption that the seeded shipping methods serve the US zone. Sylius' fixtures give every shipping method a random zone, and the US is alone in its zone, so on a fixture load where every enabled method lands in "Rest of the World" nothing ships to the US and both journeys time out on the shipping step. The checkout now tries the countries the channel offers in turn until the shipping step lists a method (it checks for the method choices, not the form's hidden token, which is rendered even when nothing ships).

Checked by moving the single enabled method (`ups`) into `WORLD` in a seeded database: the old helper times out in both journeys, the new one passes; with the zones as seeded it passes too, and the full suite passes (50). #342 and #340 are rebased onto this commit; for them it is the only change.

PR #339, opened 2026-09-14
Fit the gift card message to the card and keep its line breaks

### LINKED ISSUE #314: [Bug]: The 500-character message limit overflows the card, and line breaks are dropped

loevgaard:
## Summary

The shop form let a customer write 500 characters with no hint and no counter, and the card dropped every
newline they typed — the message was rendered into a `<div>` with no `nl2br`/`white-space`, and the live
preview wrote it with `textContent`. The limit is now configurable as
`setono_sylius_gift_card.purchase.maximum_message_length` (default 200, minimum 1) and drives the textarea's
`maxlength`, a help text and a remaining-characters counter under the field. The card renders the message with
`white-space: pre-line`, which dompdf 3.x implements, so the PDF and the preview break the lines the same way.
Because both message blocks grow upwards from a bottom anchor, preserving newlines makes a message of many
short lines lift the 52px title into the brand row, so the message is clamped to four lines; dompdf turns
`overflow: hidden` into a PDF clipping rectangle, so the clamp holds on the printed card too. The admin form
is held to the same budget through a `GiftCardMessageLength` constraint on the model, whose validator reads the configured limit.

Fixes #314

## Changes

- `purchase.maximum_message_length` in `Configuration` (default 200, `min(1)`), exposed as the container
  parameter `setono_sylius_gift_card.purchase.maximum_message_length` and injected into
  `GiftCardInformationType`, which uses it for the `maxlength` attribute and the help text in place
  of the removed 500-character constant.
- A `GiftCardMessageLength` constraint (validator built with `%setono_sylius_gift_card.purchase.maximum_message_length%`) on `GiftCard::$customMessage` in `validation/GiftCard.xml` and on `GiftCardInformation` in the new `validation/GiftCardInformation.xml`, which also carries the amount constraints the shop form used to declare inline.
  (group `setono_sylius_gift_card`), so the admin form cannot store a message the card has no room for.
- A remaining-characters counter under the textarea, rendered by
  `shop/product/show/_gift_card_information.html.twig` (so it is already correct before the script runs) and
  updated by `product-gift-card.js`; a help text naming the limit. New keys `ui.characters_left` and
  `form.gift_card_information.custom_message_help` in `en`, `da` and `fr`.
- `white-space: pre-line`, `overflow-wrap: break-word`, `line-height: 1.4` and a four-line clamp
  (`max-height: 5.6em; overflow: hidden`) on `.ssgc-card__message` / `.ssgc-card__scrim-message` in
  `shop/gift_card/_card_style.html.twig`, which is shared by the PDF and the preview.
- Tests: `ConfigurationTest` covers the new default, an override and the `min(1)` rejection;
  `GiftCardPdfGeneratorTest` gained a small PDF text-extraction helper plus two cases (line breaks survive
  into the PDF, and a 200-character 67-line message keeps the title clear of the brand row on two pages);
  two Playwright cases in `shop/gift-card-product.spec.js` for the counter, the preserved line break in the
  preview, and the clamp.
- README documents the setting; both forms follow it. Configuration refuses a value above what the TEXT column holds (65535).

## Verification

All run from the worktree, against `origin/1.x` (the branch was rebased onto `548a9ac` after 1.x moved).

- `vendor/bin/phpunit --testsuite unit` — OK (57 tests, 296 assertions)
- `vendor/bin/phpunit --testsuite functional` — OK (19 tests, 82 assertions), own database `ssgc_message_length_*`
- `bin/console lint:twig ../../src/Resources` — `[OK] All 19 Twig files contain valid syntax`
- `bin/console lint:container` — `[OK] The container was linted successfully`
- Playwright, full suite against a served `tests/Application` (`symfony serve --port=8117 --no-tls`, fresh
  schema + fixtures) — **29 passed**, including the two new shop cases.

Regression proof (the guide's step 7, done with `git checkout <base> -- <paths>`, never `git stash`):

- With the base `_card_style.html.twig`, `it_keeps_the_line_breaks_the_customer_typed` fails
  (`Failed asserting that an array contains 'Happy birthday!'` — the two sentences arrive as one text run).
- With `white-space: pre-line` but the clamp removed,
  `it_does_not_let_a_message_of_many_lines_push_the_title_off_the_card` fails
  (`Failed asserting that 1186.901 is less than 1147.901` — the title has been pushed off the page).
- With the base templates/JS/CSS, both new Playwright cases fail (no counter element, no clamp).

## Notes for reviewers

- **The admin limit is a literal 200 in `GiftCard.xml`, not the parameter.** A validation mapping cannot read
  a container parameter, and the alternative — a bespoke constraint plus validator plus service that only
  wraps `Length` — is a lot of machinery for one integer. The mapping therefore mirrors the *default* of
  `maximum_message_length`, the XML says so, and the README tells a shop that raises the setting to override
  the mapping with the same value (the standard Sylius/Symfony validation override). Happy to swap it for a
  parameter-reading constraint if you would rather the two could never drift.
- **Why four lines, and why `max-height` rather than a font step-down.** I measured dompdf 3.1.6 rather than
  guessing: it implements `white-space: pre-line`, clamps a block to `max-height`, and emits a real PDF
  clipping rectangle for `overflow: hidden`, so the same declaration works in both engines. Four lines
  (`4 x 1.4em`) is what the card can hold — a 200-character message with no hard breaks wraps to three lines
  and is never clipped, and the worst case (67 two-character lines) leaves the title ~55pt clear of the brand
  row. A font step-down would have needed a length-derived class in `_card.html.twig`, which is touched by
  #295/#312.
- **Touched by in-flight PRs.** `GiftCardInformationType.php` (#282, #313) — only the `customMessage` hunk is
  edited here. `_card_style.html.twig` (#312, #313) — one appended rule. `_card.html.twig` (#295, #312) is
  **not** touched at all. `README.md` and `SetonoSyliusGiftCardExtension.php` also moved on 1.x; the rebase
  onto `548a9ac` was clean.
- **Pre-existing, not fixed here:** in the *live preview* only, the `VALUE` label of the meta row sits higher
  than the design intends (the shop theme's inherited `line-height` inflates those line boxes), so it visually
  overlaps the last line of the message — with a one-word message too, and on `1.x` as well. The PDF is
  unaffected because dompdf does not see the shop stylesheet. It is a separate layout bug and out of scope for
  this issue; worth its own issue if you agree.


src/Resources/config/validation/GiftCard.xml
```diff
@@ -36,6 +36,16 @@
...
+        <property name="customMessage">
+            <!-- The card clamps the message to four lines, so the admin form is held to the same budget as the shop
+                 form. This mirrors the default of setono_sylius_gift_card.purchase.maximum_message_length: a
+                 validation mapping cannot read a container parameter, so a shop that raises that setting overrides
+                 this mapping with the same value -->
+            <constraint name="Length">
+                <option name="max">200</option>
+                <option name="groups">setono_sylius_gift_card</option>
+            </constraint>
+        </property>
```
loevgaard, line 56: Maybe create a custom validator that takes the parameter as constructor argument?

src/DependencyInjection/Configuration.php
```diff
@@ -60,6 +60,11 @@ public function getConfigTreeBuilder(): TreeBuilder
                             ->defaultNull()
                             ->min(1)
                         ->end()
+                        ->integerNode('maximum_message_length')
+                            ->info('The maximum number of characters a customer may write on a gift card')
+                            ->defaultValue(200)
+                            ->min(1)
+                        ->end()
```
loevgaard, line 68: What's the limit on the actual Doctrine entity?

src/Form/Type/GiftCardInformationType.php
```diff
@@ -54,13 +53,15 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
...
+                'help_translation_parameters' => ['%limit%' => $this->maximumMessageLength],
                 'required' => false,
                 'attr' => [
-                    'maxlength' => self::MAX_MESSAGE_LENGTH,
+                    'maxlength' => $this->maximumMessageLength,
                     'placeholder' => 'setono_sylius_gift_card.form.gift_card_information.custom_message_placeholder',
                 ],
                 'constraints' => [
-                    new Length(['max' => self::MAX_MESSAGE_LENGTH, 'groups' => self::VALIDATION_GROUPS]),
+                    new Length(['max' => $this->maximumMessageLength, 'groups' => self::VALIDATION_GROUPS]),
```
loevgaard: The constraints should be added in XML not here?

src/Resources/config/validation/GiftCard.xml
loevgaard: Done: a `GiftCardMessageLength` constraint whose validator receives `%setono_sylius_gift_card.purchase.maximum_message_length%` in its constructor (services/validator.xml). `validation/GiftCard.xml` uses it instead of the literal `Length`, so the admin form follows the configured limit too.

src/DependencyInjection/Configuration.php
loevgaard: None in practice: `customMessage` is mapped as `type="text"` (nullable), so the schema itself only stops at what a TEXT column holds (65,535 bytes on MySQL). The configured limit is a print/UX limit; the config node now also caps it at 65535 so a value the column cannot hold is refused at compile time.

src/Form/Type/GiftCardInformationType.php
loevgaard: Yes. The amount constraints (`NotBlank`, `ValidGiftCardAmount`) and the new `GiftCardMessageLength` now live in `validation/GiftCardInformation.xml` for `Order\GiftCardInformation`; the form only keeps the `maxlength` attribute and the help text. The one constraint still in the form is `NotBlank` on `design`, which is only added when the channel has designs at all (#325), so it depends on runtime state a mapping cannot express.

src/Validator/Constraints/GiftCardMessageLengthValidator.php
```diff
@@ -0,0 +1,41 @@
...
+
+        if (null === $value || '' === $value) {
+            return;
+        }
+
+        if (!is_string($value)) {
+            throw new UnexpectedValueException($value, 'string');
+        }
+
+        if (mb_strlen($value) <= $this->maximumLength) {
```
loevgaard: The browser and this validator count line breaks differently. A textarea's `maxlength` counts a line break as one character, but the form submits it as CR LF, and `mb_strlen()` counts two. I checked it in Chromium: with `maxlength="10"` the browser accepted `abc`, Enter, `def`, Enter, `gh` as ten characters, and PHP received twelve.

So a message near the limit that has line breaks passes in the browser and is then rejected as too long, and the live counter promises characters the server does not grant. Normalizing `\r\n` to `\n` on the way in, before validating and storing, fixes it.

loevgaard (commented):
Reviewed after merge. One bug, inline.

PR #340, opened 2026-09-14
Rate limit gift card redemption attempts and hide why a code was rejected

### LINKED ISSUE #297: [Security]: Gift card code entry has no rate limiting and the error messages reveal whether a code exists

### What happened?
- `AddGiftCardToOrderAction` (`src/Controller/Action/AddGiftCardToOrderAction.php`) has no throttling; CSRF only prevents cross-site submission, not scripted guessing by the attacker's own session.
- The responses distinguish "does not exist" (`AddGiftCardToOrderType.php:35` `invalid_message`) from "not enabled", "expired", "no balance", "already applied" (`src/Validator/Constraints/GiftCardIsApplicableValidator.php:36-63`), which lets a guesser confirm that a code exists even when it cannot be used.
- `Configuration.php:39` allows `code_length: 4` (31^4 ≈ 924k combinations), and admin-created codes are free text (`GiftCardType.php:42`), e.g. `SUMMER2026`.

Discussed in #45 (2019) and never implemented.

### Suggested fix
- Add a Symfony RateLimiter (`framework.rate_limiter.setono_gift_card_apply`, sliding window keyed by IP + session, e.g. 10/min) and consume it in the action (or a request subscriber for the two shop routes); on exhaustion flash a generic message and redirect. Make the limits configurable under `setono_sylius_gift_card.redemption.rate_limit`.
- Show shop users one generic message ("This gift card cannot be applied") for all not-found/not-usable cases; keep the specific reason in a log line and in the admin UI.
- Raise the `code_length` minimum (12) or document the risk; warn in the admin form when an entered code is shorter than the configured length.

### Acceptance criteria
`WebTestCase`: N+1 apply requests within the window → the last is rejected; unknown and disabled codes produce identical shop responses.

### Overlaps
Messages also touched by #316.

loevgaard:
## Summary

A gift card code is a bearer token, and the shop's apply form was a free oracle for guessing one: nothing
throttled the submissions, and the answers told a guesser apart a code that does not exist from one that
exists but happens to be disabled, expired, spent or for another channel. This adds a Symfony rate limiter in
front of `AddGiftCardToOrderAction`, on by default at 10 attempts per minute and pointed at any rate limiter of
the application's own through `setono_sylius_gift_card.redemption.rate_limiter`, and collapses every "cannot be used" answer into the single message the
plugin already had, `setono_sylius_gift_card.gift_card.could_not_be_applied`. The reason is not lost, it is
logged at info level with the code instead, so a shop owner can still answer "why does my gift card not
work" and can see a run of guesses in the log. Finally the minimum `code_length` goes from 4 (31^4, under a
million combinations) to 12, since throttling only buys time if the search space is large to begin with.

Fixes #297

## Changes

- Require `symfony/rate-limiter: ^6.4`.
- New configuration `setono_sylius_gift_card.redemption.rate_limiter`: the service id of the rate limiter
  factory to throttle with (per review: the application names a limiter it configured itself rather than the
  plugin growing its own limit/interval options). It defaults to `limiter.setono_sylius_gift_card_apply`, a
  sliding window limiter of 10 attempts per minute that `SetonoSyliusGiftCardExtension::prepend()` registers
  under `framework.rate_limiter` unconditionally, so its values can be overridden there like any other
  limiter's; `~` turns throttling off. `lock_factory` is set to `null` so the plugin does not drag in
  `symfony/lock`. `load()` aliases `setono_sylius_gift_card.redemption.rate_limiter` to the configured service,
  so a typo fails at compile time.
- `AddGiftCardToOrderAction` consumes a token before touching the cart or the form, and on refusal flashes
  `setono_sylius_gift_card.gift_card.too_many_attempts` and redirects without changing anything. The limiter
  factory is injected through that alias with `on-invalid="null"`, so `rate_limiter: ~` simply leaves the
  action unthrottled.
- The attempt is counted against two buckets, the client IP and the session id from the request cookie: a
  guesser that throws its cookies away still runs into the IP limit, one coming through a pool of addresses
  still runs into the session limit, and ordinary customers behind the same office NAT get a bucket each.
- One answer for every unusable code: `AddGiftCardToOrderType`'s `invalid_message` and the `GiftCardIsApplicable`
  constraint now use the same generic message; the constraint keeps a specific message only for "already
  applied", which reveals nothing the customer does not already know. `GiftCardIsApplicableValidator` and
  `GiftCardToCodeDataTransformer` log the real reason at info level (the transformer bounds and strips the
  submitted code first, so a guess cannot forge log lines).
- `collectErrors()` reports only the generic message when a child failed to transform: the unknown-code case
  otherwise produces a `NotBlank` violation on top of the transformation failure, and two messages instead of
  one is itself a tell.
- `code_length` minimum raised to 12; README configuration block documents the minimum and the new node.
- Translations: `too_many_attempts` added to `flashes.{en,da,fr}.yml` and `could_not_be_applied` to
  `validators.{en,da,fr}.yml`. No existing key removed.
- Tests: new `tests/Unit/Controller/Action/AddGiftCardToOrderActionTest.php` (real form, transformer and
  validator, Prophecy for the rest, a real `RateLimiterFactory` over `InMemoryStorage`),
  new `GiftCardIsApplicableValidatorTest`, extended `ConfigurationTest`, and a Playwright spec asserting that
  unknown, disabled and spent codes get byte-identical answers in the shop (two new seeded fixtures).

## Verification

Run from the worktree, PHP 8.4, own database `ssgc_rate_limit_*` and port 8112.

- `vendor/bin/phpunit --testsuite unit` — OK (131 tests, 462 assertions)
- `vendor/bin/phpunit --testsuite functional` — OK (24 tests, 113 assertions)
- `bin/console lint:twig ../../src/Resources` — `[OK] All 18 Twig files contain valid syntax`
- `bin/console lint:container` — `[OK]`, both with the default limiter and with
  `redemption.rate_limiter: ~` (the action then gets `null` and does not throttle); `debug:config framework
  rate_limiter` shows the test application's `limit: 100` override merged into the prepended limiter
- Playwright against a served test application on fresh fixtures: 40 passed
- Manual end to end check with `limit: 2` and curl: the first apply answers "The gift card could not be
  applied to your order.", the next ones "Too many gift card attempts. Please wait a moment before trying
  again."
- Regression checks against the unmodified code (base restored with `git checkout 090e4ec -- src/`):
  `AddGiftCardToOrderActionTest` 4 of 5 tests fail, `GiftCardIsApplicableValidatorTest` fails, and the new
  Playwright spec fails with `Expected: "The gift card with this code does not exist" / Received: "This gift
  card has no balance left."`. All pass again with the change.

## Notes for reviewers

- **The E2E suite is configured with `limit: 100`** in `tests/Application/config/packages/setono_sylius_gift_card.yaml`.
  Playwright drives everything from one address, so the shipped default of 10 per minute would throttle the
  suite itself (and any re-run inside the window) rather than a guesser. For the same reason there is no
  Playwright spec for the throttle: exhausting the bucket would poison every later spec and every retry. The
  throttling itself is covered by the unit test and was verified by hand as described above.
- **Two buckets, not one key.** The issue suggested a single window keyed by "IP + session". That key is
  bypassed by discarding the session cookie, which is exactly what a scripted guesser does, so the attempt is
  counted against the IP and the session separately instead; both must accept. The session id is read from
  the request cookie rather than from the session object, because the session is normally not started this
  early in the request and starting it would hand out a fresh id, and a fresh budget, on every attempt.
- `GiftCardIsApplicable` lost its five reason-specific message properties in favour of one `message`. A host
  application that wants its own wording still overrides `message` (or `alreadyAppliedMessage`) in its own
  validation XML. The reason-specific translation keys are left in `validators.*.yml` untouched, since #316
  is working in the same files.
- Files also touched by other in-flight PRs: `src/Resources/translations/*` (#316) and
  `SetonoSyliusGiftCardExtension::prepend()`, where the new `framework` block is deliberately appended at the
  end of the prepended array to stay out of the way.
- `composer.lock` is gitignored, so CI resolves `symfony/rate-limiter` itself. In this worktree the package
  was installed on its own to avoid re-resolving the whole dependency tree.
- Follow-up worth considering, deliberately not done here: `GiftCardToCodeDataTransformer` looks up the
  submitted code verbatim while `GiftCardApplicator` normalizes it first, so a customer typing a code the way
  it is displayed (grouped) may not find it. That is a separate bug from this one.


src/DependencyInjection/SetonoSyliusGiftCardExtension.php
```diff
@@ -433,8 +403,47 @@ public function prepend(ContainerBuilder $container): void
...
+                            'interval' => $rateLimit['interval'],
+                            // Locking would require symfony/lock, which a Sylius application does not
+                            // necessarily have, and the race it prevents only ever lets a handful of extra
+                            // attempts through, which does not matter for throttling code guesses
+                            'lock_factory' => null,
+                        ],
+                    ],
+                ],
+            ];
+        }
```
loevgaard: Wouldn't it be easier if the user could set the rate limiter to a service he already configured himself?

loevgaard: Agreed, done. The `rate_limit` block (enabled/limit/interval) is gone; there is now one option, `redemption.rate_limiter`, taking the service id of a rate limiter factory. It defaults to `limiter.setono_sylius_gift_card_apply`, a limiter the plugin still prepends under `framework.rate_limiter` (10 per minute, sliding window) so throttling is on out of the box, and the application can override those values there like any other limiter's, point the option at a limiter of its own, or set it to `~` to turn it off. `load()` aliases `setono_sylius_gift_card.redemption.rate_limiter` to whatever is configured and the action takes that alias with `on-invalid="null"`. If you would rather have no default limiter at all (opt-in only), that is one array to delete from `prepend()`.

src/Controller/Action/AddGiftCardToOrderAction.php
```diff
@@ -45,36 +60,111 @@ public function __invoke(Request $request): Response
...
+    /**
+     * The attempt is counted against the visitor's IP and against their session separately: a guesser that
+     * throws its cookies away to get a fresh session still runs into the limit for its IP, and one that comes
+     * in through a pool of addresses still runs into the limit for its session
+     *
+     * @return list<string>
+     */
+    private function rateLimiterKeys(Request $request): array
+    {
+        $keys = [sprintf('ip-%s', $request->getClientIp() ?? 'unknown')];
```
loevgaard: Would it make sense to check if the `$request->getClientIp()` is not null before adding it to the array?

src/DependencyInjection/SetonoSyliusGiftCardExtension.php
```diff
@@ -461,6 +473,26 @@ public function prepend(ContainerBuilder $container): void
             ],
         ];
 
+        // The limiter the redemption.rate_limiter option points at unless the application names another one. It
+        // is registered whether or not it is used, because the plugin's own configuration is not processed yet
+        // when prepending, and its values can be overridden under framework.rate_limiter like any other limiter's
+        $configuration['framework'] = [
+            'rate_limiter' => [
+                'enabled' => true,
```
loevgaard: This line is not needed I guess

src/Controller/Action/AddGiftCardToOrderAction.php
loevgaard: Yes, done: an unknown address gets no bucket, so such a request is counted against its session alone rather than against one bucket shared by everyone without an address. Covered by `it_counts_a_request_without_an_address_against_its_session_alone`.

src/DependencyInjection/SetonoSyliusGiftCardExtension.php
loevgaard: Right, `rate_limiter` is `canBeEnabled()` and naming a limiter turns it on; removed. `debug:config framework rate_limiter` still shows `enabled: true` with the limiter in place.

loevgaard (commented):
Reviewed at c13c6c1. The throttle and the single answer do what #297 asks. Findings inline, plus:

- **It conflicts with #338**, which rewrites `GiftCardIsApplicableValidator` around `GiftCardEligibilityChecker`. The natural merge logs the checker's `GiftCardIneligibilityReason` and adds the generic message, which is simpler than the per-branch reasons here.
- The follow-up in the description about `GiftCardToCodeDataTransformer` looking codes up verbatim is out of date. It has normalized them since #326.

src/Controller/Action/AddGiftCardToOrderAction.php
```diff
@@ -45,36 +60,118 @@ public function __invoke(Request $request): Response
...
+        if (null === $this->rateLimiterFactory) {
+            return true;
+        }
+
+        $accepted = true;
+
+        foreach ($this->rateLimiterKeys($request) as $key) {
+            // every bucket is consumed even when an earlier one already refused, so that exhausting one
+            // bucket cannot be used to spend from another
+            $accepted = $this->rateLimiterFactory->create($key)->consume()->isAccepted() && $accepted;
```
loevgaard: Both buckets have to accept, so the IP bucket is shared by everyone behind one address: an office NAT, a mobile carrier's CGNAT, or every customer of a shop behind a proxy or load balancer without `framework.trusted_proxies`. Ten attempts a minute is then the budget for all of them together. The description says customers behind the same NAT get a bucket each, which is not what happens.

At the least the README should say trusted proxies must be configured. A larger budget for the IP bucket than for the session bucket would also soften it.

src/Validator/Constraints/GiftCardIsApplicableValidator.php
```diff
@@ -64,17 +67,32 @@ public function validate(mixed $value, Constraint $constraint): void
...
         }
     }
 
+    /**
+     * Adds the generic violation the customer gets to see and keeps the actual reason in the log, where it
+     * helps the shop owner without telling a code guesser that they hit an existing code
+     */
+    private function reject(GiftCardInterface $giftCard, GiftCardIsApplicable $constraint, string $reason): void
+    {
+        $this->logger->info(sprintf(
```
loevgaard: This writes the full code of an existing card to the log, including live cards from another channel or currency. A code is a bearer token, and logs travel (log shippers, Sentry breadcrumbs), so masking it, for example keeping the last four characters, keeps the log useful without turning it into a list of spendable codes.

`GiftCardToCodeDataTransformer` has the same issue: it cannot find a disabled card or one from another channel either, so it logs those real codes in full as well.

src/DependencyInjection/Configuration.php
```diff
@@ -34,9 +34,12 @@ public function getConfigTreeBuilder(): TreeBuilder
             ->addDefaultsIfNotSet()
             ->children()
                 ->integerNode('code_length')
+                    // A gift card code is a bearer token: anyone who knows it can spend the balance. The
+                    // minimum keeps the search space out of reach of guessing (31^12 combinations), even
+                    // though applying a code is rate limited as well
                     ->info('The number of significant characters in a generated gift card code (excluding group separators)')
                     ->defaultValue(16)
-                    ->min(4)
+                    ->min(12)
```
loevgaard: An application configuring `code_length` below 12 now fails to compile, and one overriding the removed `GiftCardIsApplicable` message properties in its validation XML gets an unknown option error. Both deserve a line in UPGRADE-1.0.md.

src/Controller/Action/AddGiftCardToOrderAction.php
loevgaard: Right: both buckets have to accept, so the IP bucket was the budget for everyone behind the address, and the description said the opposite. Corrected there.

In 889d521 the IP bucket gets its own option, `redemption.ip_rate_limiter`, next to `redemption.rate_limiter`, which now throttles per session alone. It has the same shape as the first: the service id of a rate limiter factory, defaulting to a limiter the plugin prepends (`setono_sylius_gift_card_apply_ip`, sliding window, 50 attempts a minute) whose values can be overridden under `framework.rate_limiter`, and `~` turns that bucket off. Five times the per visitor budget is the ratio Symfony's login throttling keeps between its per IP and per username limiters (`LoginThrottlingFactory`). I chose a second flat option over something like `rate_limiter: {session: …, ip: …}` because it keeps `rate_limiter` the plain service id you asked for, and each bucket can be switched off on its own, which is the way out for a shop behind a proxy it cannot list in `trusted_proxies`: `ip_rate_limiter: ~`, and the session bucket still applies.

The README has a new "Protecting codes from guessing" subsection under Configuration: behind a reverse proxy or load balancer, `framework.trusted_proxies` must be configured, because otherwise Symfony takes the proxy's address for every client, all customers share one IP bucket, and once its 50 attempts a minute are used up everybody who tries a code is told to wait, customers with a real card included. UPGRADE-1.0.md points at it as well.

Covered in `AddGiftCardToOrderActionTest`: a second visitor on the same address gets through after the first used up their own budget, fresh sessions are refused once the address is exhausted, and each bucket can be turned off on its own. Checked by hand with curl and the limits lowered to 2 and 4: session A is answered twice and then refused, session B on the same address is answered once and then refused when the address runs out, and a fresh session C is refused straight away.

src/Validator/Constraints/GiftCardIsApplicableValidator.php
loevgaard: Done in 9c4a029. `GiftCardCodeNormalizerInterface` gets `mask()` next to `normalize()` and `format()`, since it is the one place that knows what a code looks like. It masks the canonical code down to its last four characters (`************MNOP`), but never more than a third of it, because an admin can type a short code of their own (`SUMMER26` becomes `******26`). Normalizing first also strips control characters, and anything longer than 255 characters (longer than any code can be) is cut, so this replaces the transformer's own sanitizing. The validator (now #338's `GiftCardIsEligibleValidator`, after the restack) gets the normalizer through its interface alias, and `GiftCardToCodeDataTransformer` already had it. Tests in `GiftCardCodeNormalizerTest`, plus a case each in the validator and transformer tests asserting that the log line carries the masked code and not the real one; both fail when the masking is taken out. After the Playwright run the plugin's lines in `dev.log` read `No enabled gift card with the code "***********0001" exists in the channel "FASHION_WEB"` and `The gift card "**********CE01" was not applied to the order: no_balance`.

Looking for other places a code of an existing card reaches a log: `InsufficientGiftCardBalanceException` and `GiftCardCurrencyMismatchException` put it in their message, which Symfony logs and error trackers pick up when one goes uncaught, so in 2387ce7 they name the card by its id instead. `GiftCardNotFoundException` keeps its code, since no card has it.

Two I left alone. Symfony's router logs route parameters at info level (`Matched route …`), and the remove route carries the code in its path (`/gift-cards/{giftCard}/remove`), so removing a card from the cart writes that customer's code to the application log, and to the web server's access log. Taking the code out of the URL changes the route, the cart template and the CSRF token id, which I would rather do in its own PR if you agree. The other is Doctrine's SQL logging of query parameters, which only happens at debug level.

src/DependencyInjection/Configuration.php
loevgaard: Added in 2878b6f. Under "Configuration migration": `code_length` must be at least 12 (0.12.x accepted anything from 1 and defaulted to 20), with the compile error a shorter value produces (`The value 8 is too small for path "setono_sylius_gift_card.code_length". Should be greater than or equal to 12`), and that codes already issued keep working whatever their length. Next to it, a line that applying a code is now rate limited and that `framework.trusted_proxies` needs configuring behind a proxy.

A new "Validation overrides" section says the constraint is `GiftCardIsEligible` (called `GiftCardIsApplicable` earlier in the 1.x development), that its only options are `message` and `alreadyAppliedMessage`, that validation XML still setting one of the removed ones fails with `The options "notEnabledMessage" do not exist in constraint "…\GiftCardIsEligible"`, and to set `message` instead.

loevgaard:
On the review body:

**Restacked onto #338.** The branch is rebased onto `fix/stale-coverage` (699d1af) and the PR now targets it; merge #338 first, and GitHub retargets this one to `1.x` when #338 merges.

The conflict came out as you suggested. The validator is #338's `GiftCardIsEligibleValidator(CartContextInterface, GiftCardEligibilityCheckerInterface)`, with the code normalizer and the logger after the checker. The per-branch reasons this PR had are gone: when the checker returns a `GiftCardIneligibilityReason`, the validator logs `$reason->value` together with the masked code and adds the constraint's one `message`; otherwise it keeps #338's already-applied check. The `match` over the reason-specific messages that #338 added went with those properties, and `GiftCardIneligibilityReason`'s docblock, which said the cases mirror the messages the constraint renders, now says the customer never learns which one it was. `GiftCardIsApplicableValidatorTest` became `GiftCardIsEligibleValidatorTest` and runs every reason through a prophesized checker. The redemption spec takes `GIFT_CARD_FIELD`, `addSomethingToCart` and `applyGiftCard` from `specs/support/cart.js`. The wiring follows #352: the validator gets the checker and the normalizer through their interface aliases, and `InterfaceAliasWiringTest` passes.

**The stale follow-up.** You are right, `GiftCardToCodeDataTransformer` has normalized codes since #326; the note is removed from the description.

Verification after the restack and the review commits: unit OK (199 tests), functional OK (48 tests), `composer analyse` and `check-style` clean, rector clean, `lint:container` OK in dev and test (also with either limiter set to `~`), and the full Playwright suite on fresh fixtures 51 passed. One thing from that run that belongs to #338 rather than here: its `stale-gift-card.spec.js` picks `US` as the address country on the assumption that the seeded shipping methods serve the US zone, but Sylius' fixtures give each shipping method a random zone. On my first fixture load every enabled method landed in `WORLD` and both stale card journeys timed out on the shipping step; after reloading the fixtures they passed.

loevgaard:
Re-reviewed: every point is addressed. Rebasing onto 1.x after #342 and #343 conflicted only in the `load()` config shape docblock, once per commit that reshapes `redemption`; each time I kept #343's `delivery` line and this PR's `redemption` shape. Unit, PHPStan and ECS were clean locally, and it was merged once all 26 checks were green.

PR #341, opened 2026-09-14
Draw a real barcode and keep the redemption code on the gift card PDF back

### LINKED ISSUE #302: [Bug]: PDF back side shows a fake barcode, promises in-store redemption, and hides the code when a back image is set

### What happened?
- The barcode on the PDF back (`src/Resources/views/shop/gift_card/pdf.html.twig:6`, drawn at `:60-62`) is a hardcoded array of bar widths, identical on every card, encoding nothing. Next to it the copy says "present this card in store" (`messages.en.yml:92`, `pdf.how_to_redeem_text`). A clerk will try to scan it.
- When a design has a back image, the back is **only** the image (`:47-48`): no code, no redemption text. The front scrim still shows the (dashed, see #287) code.
- The layout is absolute-px "tuned for ~560x396" (`_card_style.html.twig:4`), i.e. A6 landscape, while `pdf.page_size` invites A4/letter which do not scale.

### Suggested fix
- Render a real Code 128 (or QR) of the code, e.g. `picqer/php-barcode-generator` as an inline SVG/PNG data URI, or drop the barcode entirely. Decide which by whether POS redemption is a target (#274 asks for a public balance page; a QR to it would be a natural use).
- Reword `pdf.how_to_redeem_text` to online redemption ("enter the code at checkout on {hostname}") unless in-store is supported.
- With a back image, overlay the code box on top of the image.
- Either scale the card to the page (CSS `@page` size + `vw`-based layout) or restrict `page_size` to what the layout supports.

### Overlaps
`pdf.html.twig` also touched by #287, #295; copy by #316.

loevgaard:
## Summary

The back of the gift card PDF was making promises it could not keep. The barcode was a hardcoded array of bar widths, the same drawing on every card, encoding nothing, and the copy next to it invited the customer to present the card in a store, which nothing in the plugin supports: a gift card is redeemed in the shop's own checkout, where it becomes a payment on the order. When a design had a back image the back was *only* that image: no code, no redemption copy, no terms, so the one thing the back of a gift card exists for disappeared exactly when a merchant bothered to upload artwork. And the layout is a fixed pixel grid tuned for A6 landscape, while `pdf.page_size` invites A4 or letter, on which the card was drawn at its literal size into the corner of a mostly empty sheet.

The barcode is gone (per review: a host that wants one, or a QR code, renders it in its own template override), the copy sends the customer to the shop's own checkout by hostname, the code and the copy are overlaid on a back image instead of being replaced by it, and the card scales onto whatever page size is configured.

Fixes #302

## Changes

- **No barcode.** The fake one is removed from the template and the stylesheet along with the `barcode` block, and the README no longer mentions it. The card back now holds the redemption code, the copy and the terms.
- **Code over artwork.** A back image now keeps a band at the top of the card and the rest of the back sits on a panel over it, so the redemption code, the copy and the terms read the same way whether or not a design brings an image. One set of positions serves both variants; the only difference is the panel.
- **Online redemption copy.** `pdf.how_to_redeem_text` keeps its key in all three locales and takes a `%hostname%` placeholder, filled from the channel hostname (falling back to the channel name). No other translation key is touched (#316).
- **Page size.** `DompdfGiftCardPdfGenerator` works out the factor between the 560 px design grid and the paper width and the template scales each side onto the page with a CSS `transform: scale()`. Each side is an absolutely positioned box inside a page-sized wrapper (`.page`, with `page-break-before` on the back), so a box taller than the page, which it is on every page narrower than the design (A6 itself, by a hair), is not split over two pages before the transform shrinks it. Paper of a different aspect ratio gains or loses height, which the layout absorbs: every block is anchored to the top or the bottom of the card, never sized against another. The pixel values in `_card_style.html.twig` stay the single definition of the layout; a comment there says so.
  - Not dompdf's `dpi` option, which an earlier revision of this PR used: dompdf 2.x caches pixel-to-point conversions per process without the dpi in the cache key (`Style::length_in_pt()`; 3.x keys on it), so the first dpi used in a process silently wins for every later render. That is what made the lowest-deps functional job fail, and it would bite any long-running worker that rendered PDFs at another dpi first.
- **Tests.** Functional tests for the overlay with and without a back image, the hostname copy and the scaling; the scaling test reads text positions through the content stream's transformation matrix (`q`/`Q`/`cm`), so it measures where text lands on the page rather than how it got there, and passes on dompdf 2.0.4 and 3.1.6. The Playwright PDF specs inflate the PDF's content streams and assert the card actually prints the code the admin shows.
- The seeded `classic` design gets a back image, so the design preview spec covers the case that used to drop the code. The README documents the scaling.

## Verification

Run from the worktree on PHP 8.4 after rebasing onto the current `1.x`, with the highest dependency set (dompdf 3.1.6) and again with `composer update --prefer-lowest` (dompdf 2.0.4) for the functional suite.

| Command | Result |
| --- | --- |
| `vendor/bin/phpunit --testsuite unit` | OK (120 tests, 436 assertions) |
| `vendor/bin/phpunit --testsuite functional` | OK (27 tests, 132 assertions), highest and lowest |
| `composer analyse` | `[OK] No errors` |
| `composer check-style` | `[OK] No errors found` |
| `vendor/bin/rector process --dry-run` | `[OK] Rector is done!` |
| `bin/console lint:twig ../../src/Resources` | `[OK]` |
| `npx playwright test` (served app, own port and database, fresh fixtures) | 43 passed |

The dompdf 2 cache was pinned down by generating A6, A5, A4, letter and A7 PDFs in one process and reading the content streams: with the `dpi` approach the second size rendered inherits the first one's conversion; with the transform every size comes out with the text at 7.5 % of the page width and two pages.

## Notes for reviewers

- **The overlay covers about the top third of the artwork only.** Making the panel smaller would mean a second set of positions for the image variant, which is how the two backs drifted apart in the first place. If the artwork matters more than the terms on that side, the panel is one CSS rule to shrink.
- **`transform` in dompdf** is applied to the frame and everything inside it (`Renderer::render()` wraps the subtree in `save()`/`restore()` around the canvas transform), which is why the whole side scales together, images included. dompdf does not support viewport units, so the `vw` layout the issue suggests is not available.
- **`src/Resources/config/app/fixtures.yaml`** now gives the seeded design a back image. That changes what a host's `sylius:fixtures:load` demo data looks like, which seemed the right trade for having the suite exercise the case the issue is about.
- Files also touched by in-flight work: `pdf.html.twig` and the translations sit under #287 and #295 (both already in this branch's base) and the copy under #316. #343 stacks on this branch but only touches the email templates and the operator.


loevgaard:
I don't want the barcode. If plugin users want to render barcodes, they can do it themselves

loevgaard:
Barcode dropped: the generator, its interface and unit test, the Twig filter, the services and the template/stylesheet block are gone, and the README no longer mentions it. A host that wants a barcode or QR code can render one in its own override of `pdf.html.twig`.

While making the lowest-deps build green I also had to replace the dpi-based page scaling: dompdf 2.x caches pixel-to-point conversions per process without the dpi in the cache key, so the first dpi used in a process silently wins for every later render (that is what the failing `it_scales_the_card_onto_the_configured_page_size` on PHP 8.x / lowest was showing). The card is now scaled onto the page with a CSS transform instead; details in the updated description.

loevgaard (commented):
Reviewed after merge. The code is fine. The title still says "Draw a real barcode", but the barcode was dropped per review, which is worth knowing when reading the history. It also matters for #343, which still carries the pre-review barcode commit.

Issue #291, opened 2026-09-11
[Docs]: README omits required install steps and undocumented runtime behaviour


loevgaard:
### Scope
Documentation only (README.md, UPGRADE-1.0.md, UPGRADE-0.12.md). Everything below was verified against `tests/Application` and the code at 090e4ec.

### Missing or wrong
1. **Customer repository override is required and undocumented.** `src/Resources/config/routes/admin_ajax.yaml:5` routes the admin customer autocomplete to repository method `findByEmailPartForGiftCard`, which exists only in `src/Doctrine/ORM/CustomerRepositoryTrait.php`. The reference app applies it (`tests/Application/Doctrine/ORM/CustomerRepository.php`, `tests/Application/config/packages/_sylius.yaml:17-21`). README never mentions `CustomerRepositoryTrait` / `Repository\CustomerRepositoryInterface` and gives no `_sylius.yaml` YAML at all (`README.md:143` just points at tests/Application).
2. **`OrderRepositoryTrait` has no consumer in `src/`** (`findLatestByCustomer`, `createQueryBuilderByGiftCard`, `createCompletedQueryBuilderByGiftCard`) yet the reference app applies it and `phpstan.neon:40-47` suppresses `trait.unused` calling it public API. Drop it or document what it is for.
3. **LiipImagineBundle is a hard requirement** (`imagine_filter` in `src/Resources/views/shop/product/show/_gift_card_information.html.twig:40-42`, filter sets prepended in the extension) – not stated.
4. **Fixtures are not prepended**: `src/Resources/config/app/fixtures.yaml` must be imported manually (`tests/Application/config/packages/setono_sylius_gift_card.yaml:1-2`), contradicting `README.md:73`. Also note that fixture `amount` is in major units (`GiftCardFixture.php:24` float, ×100 in the example factory) while `purchase.minimum_amount` is minor units.
5. **UI block ids** (`setono_gift_card_information`, `setono_gift_cards`, `setono_gift_card_totals` on `sylius.shop.product.show.add_to_cart_form`, `sylius.shop.cart.summary`, `sylius.shop.cart.summary.totals`) are not listed, so hosts cannot disable or re-prioritise them.
6. **Assets are injected inline** (`<script defer>` + `<link>` inside the product form block, `_gift_card_information.html.twig:52-53`), not Encore entries – matters for CSP/nonce setups.
[...]

PR #342, opened 2026-09-14
Restore the balance when a gift card payment is refunded and disable the cards a refunded order bought

loevgaard:
Stacked on #338 (`fix/stale-coverage`, issue #278): this branch was created from that branch and this PR targets it, so the diff shown here is only the refund fix. Merge #338 first, then this one.

## Summary

Only `sylius_order.cancel` was hooked, so the README's "restored when it is cancelled/refunded" was half true: an admin refunding a gift card payment (the admin's Refund button applies `sylius_payment.refund`) left the card drained with no ledger row, and refunding an order that bought gift cards left the cards spendable on top of the money that went back. The `refund` transition of the payment graph now restores the balance on both state machine adapters, keyed on the payment alone (`restore:payment:<id>`) with the ledger row linking the payment, and it is the one place the balance comes back from: `rollback()` on cancel now only refunds the order's gift card payments. A full refund of an order (`sylius_order_payment.refund`) disables the cards it bought; `partially_refund` deliberately does nothing, because it does not say which payments or items the money went back for, so the cards stay usable and the merchant disables them by hand where that is what the refund meant.

Fixes #284

## Changes

- `GiftCardRedemptionMethodInterface::rollbackPayment(PaymentInterface)` (new), implemented in `PaymentRedemptionMethod`: for a gift card payment in state `refunded`, restores `payment->getAmount()` to the card via the balance operator with key `restore:payment:<id>`, order and payment linked on the ledger row. Ignores payments that are not gift card payments or are not refunded.
- `PaymentRedemptionMethod::rollback()` now only applies `refund` to each gift card payment (the `can()` check keeps it to completed ones); the refund hook does the restoring.
- winzou (`prepend()`): `sylius_payment` after `refund` -> `rollbackPayment` at -150 (before `sylius_resolve_state` at -100); `sylius_order_payment` after `refund` -> `OrderGiftCardOperator::disable` at 0 (Sylius registers nothing on that transition). Both with a comment stating where they sit.
- Symfony Workflow: new `EventSubscriber/Workflow/RollbackPaymentSubscriber` (`workflow.sylius_payment.completed.refund`, 150); `DisableGiftCardsSubscriber` additionally listens to `workflow.sylius_order_payment.completed.refund` (0). Registered in `services/workflow.xml`.
- `StateMachineCallbackParityTest` picks the event subject by graph (`sylius_payment` -> `PaymentInterface`), so subscribers on the payment graph are checked against their winzou twins too.
- Tests: `tests/Functional/GiftCardPaymentRefundTest.php` (refunding a gift card payment on a partially covered order, on winzou and through Symfony Workflow: balance back exactly once, one `restore` row linked to the payment and order, re-firing `rollbackPayment()` is a no-op, Sylius still resolves the order to `partially_refunded`; cancelling the order on both adapters still restores exactly once), `tests/Functional/PurchaseOrderRefundTest.php` (refunding the payment of an order that bought a card cascades into `sylius_order_payment.refund` and disables the card; the same through the workflow; a partial refund on both adapters leaves it enabled), Prophecy unit tests for `RollbackPaymentSubscriber`, for the refund subscription of `DisableGiftCardsSubscriber`, and for `rollbackPayment()`/`rollback()` (`tests/Unit/Redemption/PaymentRedemptionMethodTest.php`).
- README.md and CLAUDE.md: one sentence each on what is restored and disabled, and when (no doc rewrite, #291 owns that).

## Verification

- `vendor/bin/phpunit --testsuite unit` -> `OK (83 tests, 456 assertions)`
- `vendor/bin/phpunit --testsuite functional` (own database `ssgc_refund_handling_test`) -> `OK (31 tests, 189 assertions)`
- `composer analyse` -> `[OK] No errors`
- `composer check-style` -> `[OK] No errors found. Great job - your code is shiny in style!`
- `vendor/bin/rector process --dry-run` -> `[OK] Rector is done!`
- `bin/console lint:container` -> `[OK] The container was linted successfully`; `bin/console lint:yaml ../../src/Resources` -> `[OK] All 14 YAML files contain valid syntax.` (no Twig touched)
- Regression proof: with the five touched `src` files restored to `fix/stale-coverage`, `GiftCardPaymentRefundTest` fails its two refund tests (`FAILURES! Tests: 4, Assertions: 38, Failures: 2`) and `PurchaseOrderRefundTest` fails its two full-refund tests (`FAILURES! Tests: 4, Assertions: 13, Failures: 2`). The two cancel tests and the two partial-refund tests pass on the base, as they should: cancel behaviour is unchanged and a partial refund is deliberately left alone.
- Playwright: not run, there is no UI change (no template, JS, CSS or route touched).

## Notes for reviewers

- Deviation from the sketch in the issue, on purpose: the issue suggested that both `rollback()` and the refund hook restore with an aligned idempotency key. That cannot work. `GiftCardBalanceOperator` checks the key with `findOneBy()`, which queries the database, and nothing is flushed between `rollback()`'s restore and the nested `refund` transition it applies right after, so the hook would not see the first ledger row, restore again, and the unique index on `idempotencyKey` would then reject the flush. Making the refund hook the single restore path is the smallest fix that is actually correct; the payment-keyed idempotency still guards re-fires after a flush (covered by the re-fire step in the functional tests).
- `rollbackPayment()` is added to `GiftCardRedemptionMethodInterface`. 1.0.0 is not tagged yet (latest tag is `v0.12.0-beta`), so no released contract changes; an application substituting its own redemption method has one more method to implement, and a method that does not redeem through payments can leave it empty.
- The restore key changed from `restore:order:<order>:payment:<payment>` to `restore:payment:<payment>`. Rows written under the old key keep it; nothing can restore the same payment again anyway, since a payment cannot be refunded twice.
- Priorities: the restore runs at winzou -150 / workflow 150, before Sylius resolves the order's payment state from the refunded payment, so the card is whole again before anything reacts to the order being (partially) refunded. The disable on order refund sits at 0 on both adapters because Sylius registers nothing on that transition.
- `rollbackPayment()` restores only for a payment in state `refunded`, so calling it on a payment that is still completed cannot hand out money the shop kept.
- Files also touched by in-flight PRs: `prepend()`, `services/workflow.xml`, `PaymentRedemptionMethod.php` and the parity test are on the base branch (#338) and this PR builds on their state there; `services/redemption/payment.xml` and the order payment processor (#320) are not touched.
- Out of scope: a per-unit refund through the Sylius Refund Plugin could disable exactly the refunded units' cards; that would be a follow-up if the plugin ever integrates with it.


loevgaard (commented):
Reviewed the refund commit, 30e0de8. The logic is right. `refund` only leaves `completed`, so `rollback()` relying on `can()` changes nothing. The restore has a single path keyed on the payment. It runs at -150, before Sylius' `sylius_resolve_state` at -100.

**It needs a rebase first.** The branch merged the pre-review version of #338, so it still carries `Checker\GiftCardApplicabilityChecker`, `Guard\GiftCardCoverageGuard` and the subscriber in `event_listener.xml`. Merged into the current `fix/stale-coverage`, those would come back next to the renamed classes. Rebasing the single refund commit onto the current #338 branch fixes that. The failing End to end job goes with it: it runs the old `stale-gift-card.spec.js`, from before #338's checkbox fix.

loevgaard:
Rebased: the branch is now the current `fix/stale-coverage` (699d1af) plus the refund commit alone, cherry-picked as 82f0ba2. The old #338 commit (a88fce4) and the two merge commits are gone, so `git log origin/fix/stale-coverage..fix/refund-handling` shows just the refund commit.

- **Old #338 classes:** none of them are on the branch any more. `Checker\GiftCardApplicabilityChecker`, `GiftCardInapplicabilityReason`, `Guard\GiftCardCoverageGuard` and the subscriber entry in `event_listener.xml` came only from the pre-review #338 commit and the merges, and the refund commit never touched them. The diff against the base has no `Applicab`, `Guard\` or `event_listener` match.
- **Conflicts:** only the `use` blocks of `StateMachineCallbackParityTest` and `WorkflowSubscriberTest`, where #338 moved the guard interface to `StateMachine\`. I kept the renamed import next to the refund commit's `PaymentInterface`/`PaymentTransitions`. Everything else applied cleanly.
- **#352 wiring:** already in line, so the commit needed no other change and its message still describes it. The new winzou callbacks call public aliases that resolve through the interface: `@setono_sylius_gift_card.redemption_method::rollbackPayment` on `sylius_payment.refund` (-150) and `@Setono\SyliusGiftCardPlugin\Operator\OrderGiftCardOperatorInterface::disable` on `sylius_order_payment.refund` (0). `RollbackPaymentSubscriber` is built with `setono_sylius_gift_card.redemption_method`, like the commit and rollback subscribers next to it, and `DisableGiftCardsSubscriber` keeps its `OrderGiftCardOperatorInterface` argument. `InterfaceAliasWiringTest` and `StateMachineCallbackParityTest` pass with the priorities unchanged.
- **End to end:** the branch now runs #338's current `stale-gift-card.spec.js`. Locally the whole Playwright suite passed 50/50 against a freshly seeded database. On CI the End to end tests (PHP 8.3) job now passes, and so do the other 23 jobs of the build (run 35995803833) and both codecov checks.

Verification after the rebase: unit (174 tests), functional (56 tests), `composer analyse`, `composer check-style`, `rector --dry-run`, `lint:container` (dev and test) and `lint:yaml` all pass. I also re-ran the regression proof: with the refund commit's `src` changes reverted, the two refund tests in `GiftCardPaymentRefundTest` and the two full-refund tests in `PurchaseOrderRefundTest` fail as before.

loevgaard:
Re-reviewed: the rebase removed the old #338 commits and the refund commit is unchanged. Rebasing onto 1.x after #338 and #343 conflicted in README's "Buying a gift card" paragraph; I kept #343's email wording and appended this PR's sentence on cancelling and refunding. Merged once all 26 checks were green.

PR #321, opened 2026-09-14
Group displayed gift card codes through GiftCardCodeNormalizer::format()

loevgaard:
Note: the approved #331 (email and PDF locale) has been merged into this branch, so this PR now carries both changes; #328 and #341 remain stacked on it.

## Summary
Displayed gift card codes were grouped by hand-written 4-4-4-4 slicing in the PDF, cart table, both email templates and the admin show page, which silently assumes every code is exactly 16 characters. Since `code_length` is configurable (4..255) and admins can type any code in `GiftCardType`, a shorter or longer code rendered with stray dashes (e.g. `SUMMER26` as `SUMM-ER26--`) even though `GiftCardCodeNormalizer::format()` already implements correct grouping and was simply never called from templates.

This adds a `setono_gift_card_format_code` Twig filter (extension + runtime, mirroring `GiftCardRedemptionExtension`/`Runtime`) that delegates to `GiftCardCodeNormalizerInterface::format()`, and uses it at every place a code is shown. `format()` itself is switched from `str_split` to `mb_str_split` so it groups correctly for a code no longer than one group, an empty code, and a code containing a multibyte character, instead of assuming a fixed length.

## Changes
- Added `Setono\SyliusGiftCardPlugin\Twig\Extension\GiftCardCodeExtension` and `Twig\Runtime\GiftCardCodeRuntime`, registered in `src/Resources/config/services/twig.xml`; the runtime renders a `null` code as an empty string so templates don't need a null check.
- `GiftCardCodeNormalizer::format()` now uses `mb_str_split` instead of `str_split`.
- Replaced the hardcoded slicing in `shop/gift_card/pdf.html.twig` and the raw code in `shop/cart/_gift_cards.html.twig`, both `email/*.twig` templates and `admin/gift_card/show.html.twig` with the new filter. Only the displayed code changed — form actions, CSRF tokens and test-html attributes still use the raw code (see `_gift_cards.html.twig`).
- Unit tests for `GiftCardCodeNormalizer::format()` covering 8, 15, 16 and 20 character codes, a code no longer than one group, an empty code, a multibyte code, and a normalize(format(x)) == x round trip; Prophecy unit tests for `GiftCardCodeExtension`/`GiftCardCodeRuntime`.
- Functional coverage: `GiftCardPdfGeneratorTest` renders the PDF template for an 8-character admin-typed code and asserts it shows `SUMM-ER26` without a trailing dash; `GiftCardEmailManagerTest` asserts the emailed HTML shows the grouped code.
[...]

Issue #270, opened 2026-08-17
[Feature request]: Recipient name and email fields on a gift card


loevgaard:
### Describe the proposed solution

Let the buyer address a gift card to someone: capture a recipient name and email when composing it on the product page, and deliver a virtual card to that address instead of to the buyer.

Today a virtual gift card is emailed to the **buyer's** customer email, so gifting someone else means forwarding the PDF by hand.

### Additional context

Deferred during the 1.x rewrite (#269) — see the roadmap in `REWRITE.md`.

Where it would land:

- `Order/GiftCardInformation` (+ `GiftCardInformationType`) already carries the per-unit amount, message and design chosen at add-to-cart; recipient name/email belong beside them
- `Model/GiftCard` would need to persist them, snapshotted by `Operator/OrderGiftCardOperator::reconcile()` like the other fields
- `Mailer/GiftCardEmailManager::sendGiftCard()` currently resolves the address from `$giftCard->getCustomer()?->getEmail()` and would prefer the recipient address when present

Worth deciding whether a recipient address should also create/attach a Sylius customer, or stay a plain string on the gift card.

PR #343, opened 2026-09-14
Stop emailing a live code and PDF for physical gift cards

### LINKED ISSUE #303: [Bug]: Physical gift cards are emailed with a live code and the same wording as virtual ones

### What happened?
`OrderGiftCardOperator::send()` (`src/Operator/OrderGiftCardOperator.php:108-116`) emails every card of a paid order, physical ones included (documented as a "digital backup" in `GiftCardEmailManagerInterface.php:14-15`). The email body (`src/Resources/views/email/gift_cards_from_order.html.twig:6-17`) is identical for both delivery types: "Your gift card is attached to this email as a PDF" plus the live code. So a physical card the merchant then ships is redundant, and the code is spendable before it arrives. The email also omits the message, expiry date and how to redeem; `email/gift_card.html.twig` is a near duplicate of the order one.

Recipient name/email is a separate feature: #270.

### Suggested fix
- Decide the default for `Physical`: send a "your gift card is on its way" email without the code/PDF, with a config switch (`setono_sylius_gift_card.delivery.email_physical_cards: false|true`) for merchants who want the backup.
- Branch the templates on `giftCard.deliveryType.value`; include message, expiry and redemption instructions; share a partial between the two emails.
- Document the behaviour in the README.

### Acceptance criteria
Functional `GiftCardEmailManagerTest`: physical card → no PDF attachment and no code by default; virtual unchanged.

### Overlaps
Email templates also touched by #295; `send()` path by #305.

loevgaard:
Stacked on #341 — this branch is cut from `fix/pdf-back` and targets it, so please merge #341 first (which itself stacks on #331 and #321). Only the last commit belongs to this PR.

## Summary

Every gift card of a paid order was emailed the same way: the live code in the body and the card attached as a PDF. For a physical card that is wrong twice over — the card is shipped with its code printed on it, so the email duplicates what is in the envelope, and it makes the card spendable before it arrives. A physical card is now only announced as being on its way, with no code and no PDF, while a virtual card is unchanged: the email *is* the delivery, so it keeps the code and the attachment. Shops that want the digital backup anyway turn on the new `setono_sylius_gift_card.delivery.email_physical_cards`, and then a physical card is presented exactly like a virtual one.

While in there, the two email templates rendered the cards themselves in near-identical markup and told the buyer nothing but the code and the balance. Both now include one shared partial that applies the delivery-type rule in a single place and additionally prints the custom message, the expiry date and how to redeem.

## Changes

- `setono_sylius_gift_card.delivery.email_physical_cards` (bool, default `false`) added to `Configuration`, exposed as the container parameter of the same name and injected into `GiftCardEmailManager` as a third constructor argument (defaulted, so existing instantiations keep working).
- `GiftCardEmailManager::send()` skips PDF generation for a card whose code is not disclosed, and passes `emailPhysicalCards` to the templates so the body follows the same rule as the attachments. `GiftCardEmailManagerInterface`'s "always a digital backup" docblock updated to describe the new behaviour. No public signature changed, so `SendGiftCardEmailSubscriber` and `OrderGiftCardOperator::send()` are untouched.
- New `src/Resources/views/email/_gift_cards.html.twig` renders the cards for both emails; `gift_card.html.twig` and `gift_cards_from_order.html.twig` now just include it. The partial keeps the base branch's conventions: every `|trans` takes `localeCode` and the code goes through `setono_gift_card_format_code`.
- Four new keys under `setono_sylius_gift_card.email` in `messages.{en,da,fr}.yml` (`physical_on_its_way`, `physical_code_on_card`, `valid_until`, `how_to_redeem`). No existing key removed (#316).
- README: a paragraph under *Virtual vs physical* on what each delivery type is emailed, the option in the configuration block, and a pointer to the shared partial under *Customizing the emails*.

## Verification

All run from the worktree against an isolated database (`ssgc_physical_email_test`).

- `vendor/bin/phpunit --testsuite unit` — OK (71 tests, 150 assertions)
- `vendor/bin/phpunit --testsuite functional` — OK (25 tests, 114 assertions)
- `vendor/bin/phpunit tests/Functional/GiftCardEmailManagerTest.php` — OK (5 tests, 31 assertions)
- Regression check: with `src/` restored to `fix/pdf-back`, `it_does_not_email_a_physical_gift_cards_code_or_pdf` fails (`a physical gift card should not be attached as a PDF — actual size 1 matches expected size 0`); it passes with the fix.
- `composer analyse` — `[OK] No errors` (PHPStan level max)
- `bin/console lint:yaml ../../src/Resources` — OK, 14 files (caught an unquoted colon in the French string, fixed)
- `bin/console lint:twig ../../src/Resources` — OK, 19 files
- `bin/console lint:container` — OK
- `bin/console debug:translation fr @SetonoSyliusGiftCardPlugin --only-missing` — nothing missing
- No Playwright specs: nothing here is browser-rendered UI. The change is email bodies and attachments, which the functional `GiftCardEmailManagerTest` asserts directly (it captures the `MessageEvent` and reads the rendered HTML and the attachment bytes) — a browser spec could not see any of it.

Both rendered bodies were eyeballed once during development: the physical one reads "Your gift card is on its way — we are shipping it to you. The code is printed on the card itself, so you get it when the card arrives." with the balance and expiry and no code; the virtual one keeps the code, the message, the expiry, "attached as a PDF" and the redemption line.

## Notes for reviewers

- The `email_physical_cards: true` case is covered by constructing `GiftCardEmailManager` directly with the flag set, taking the sender and PDF generator from the test container. The test app has no second kernel environment to flip a container parameter in, and building one for a single boolean seemed worse than instantiating the service; the default-off path is covered through the container-wired service, and `ConfigurationTest` covers the node itself.
- The delivery-type rule lives in two places by necessity: `GiftCardEmailManager::disclosesCode()` decides the attachments, the partial decides the body. Both carry a comment pointing at the other. Putting it only in PHP would mean passing a parallel structure the template could index by card, which is worse for anyone overriding the template.
- An order can mix delivery types (one virtual and one physical card on the same order). The partial handles that: each card is presented by its own type, and the "PDF attached" / "how to redeem" lines only appear when at least one card in the email actually has an attachment.
- Files also touched by in-flight work: `src/Resources/views/email/*` come from #331 (#295) and the code formatting filter from #321 (#287) — both conventions are kept in the new partial. `OrderGiftCardOperator::send()` is also in #305's path but is unchanged here.
- Recipient name/email is #270 and deliberately not implemented; #291 owns the wider README work.


loevgaard (commented):
Reviewed the email commit, 12af51d. The email logic itself is fine. Two product questions inline.

**It needs a rebase onto 1.x first.** The branch still carries the pre-review commits of #341: "Print a real barcode and the redemption code on every gift card PDF back" with `Code128BarcodeGenerator`, and "Read the gift card PDFs back in the Playwright suite". #341 was merged without the barcode, per review, so merging this as it stands would bring the barcode back. The failing functional test on lowest deps, `it_scales_the_card_onto_the_configured_page_size`, comes from that old copy too. Only the last commit belongs here, and the description still says the PR is stacked on #341.

src/Mailer/GiftCardEmailManager.php
```diff
@@ -96,6 +106,17 @@ private function send(string $code, string $email, array $giftCards, array $data
...
 
+    /**
+     * A physical gift card is shipped with its code printed on it, so neither the code nor the PDF it is printed
+     * on belongs in the email: that would make the card spendable before it arrives, and duplicates what the
+     * customer is about to receive. Merchants who want the digital backup anyway turn on
+     * setono_sylius_gift_card.delivery.email_physical_cards
+     */
+    private function disclosesCode(GiftCardInterface $giftCard): bool
+    {
+        return $this->emailPhysicalCards || GiftCardDeliveryType::Physical !== $giftCard->getDeliveryType();
```
loevgaard: `sendGiftCard()` goes through this rule too, so the admin "Send email" action from #336 now emails a physical card without its code. That action is the one tool an admin has when a customer lost the physical card or it never arrived. Should an explicit send from the admin always disclose the code?

src/Resources/translations/messages.en.yml
```diff
@@ -89,14 +89,18 @@ setono_sylius_gift_card:
...
         redemption_code: Redemption code
         terms: 'Terms & conditions apply. This card is not redeemable for cash and cannot be replaced if lost or stolen.'
     email:
         balance: Balance
+        how_to_redeem: Enter the code at checkout to pay with your gift card. Whatever is left stays on the card for next time.
         new_gift_card: New gift card
         new_gift_card_intro: A new gift card was created for you
         pdf_attached: Your gift card is attached to this email as a PDF.
+        physical_code_on_card: The code is printed on the card itself, so you get it when the card arrives.
+        physical_on_its_way: Your gift card is on its way — we are shipping it to you.
```
loevgaard: This email goes out when the order is paid, before anything has shipped, so "we are shipping it" may be early. Wording like "will be shipped to you" fits either moment.

src/Mailer/GiftCardEmailManager.php
loevgaard: Yes. `sendGiftCard()` now always discloses: code in the body and the PDF attached, whatever the delivery type (d16a7e2). Both of its callers only run because an admin asked for the email — Send email on the show page, and the notification checkbox when creating a card (which is always virtual, the form has no delivery type) — so the rule stays only where it protects something: `sendGiftCardsFromOrder()`, the automatic email on payment, still withholds a physical card's code and PDF unless `delivery.email_physical_cards` is on.

I kept the PDF with the code on purpose. It is the card itself, which is what the customer needs in place of one that was lost or never arrived, and withholding it while sending the code would protect nothing. If the original card turns up later, both carry the same code and share one balance.

The admin email doesn't mention shipping. The order email passes `announceShipping` to the shared partial and the single card email doesn't, because a card resent from the admin may have shipped long ago. The template variable is now `disclosePhysicalCards` instead of `emailPhysicalCards`, since it is decided per email rather than copied from the option. New test `it_emails_a_physical_gift_cards_code_and_pdf_when_an_admin_sends_it` covers the default config: code, PDF, redemption line, no shipping wording. It fails without the change (0 attachments). The two existing physical-card tests now go through `sendGiftCardsFromOrder()` with an order. The README and the option's `info()` text now say the option is about the order email, and that Send email always includes the code.

One thing I left alone: the single card email still opens with "A new gift card was created for you", which reads oddly for a resend. Every resend has said that since #336, virtual cards included, so it belongs in its own change.

src/Resources/translations/messages.en.yml
loevgaard: Changed to "Your gift card will be shipped to you." (7379bb1), with the equivalent in da ("Dit gavekort bliver sendt til dig.") and fr ("Votre chèque-cadeau vous sera expédié."). Those are the only other files with the key. I renamed the key along with the text, `physical_on_its_way` → `physical_will_be_shipped`, since it only exists on this branch. After the change above, the line only appears in the order email, which is the one sent before shipping. The line after it, "The code is printed on the card itself, so you get it when the card arrives.", fits either moment as it is.

loevgaard:
Rebased onto `1.x` by rebuilding the branch as `1.x` plus the email commit only: 12af51d, now 5f66de4. #341's pre-review commits are gone: the `Code128BarcodeGenerator` barcode and the Playwright PDF read-back. So are the merge commits that carried them, so this no longer brings the barcode back. `git diff origin/1.x` only touches the email change's files, with nothing under `src/Pdf`, `src/Generator`, the PDF template or the Playwright specs. The only conflicts were in README's configuration block and in the extension's config shape and parameters, where `1.x` had added `purchase.maximum_message_length`. I kept both sides.

The lowest-deps failure of `it_scales_the_card_onto_the_configured_page_size` came from the dpi-based scaling in that stale copy of `DompdfGiftCardPdfGenerator`. The branch now has `1.x`'s CSS-transform scaling and `1.x`'s version of the test. CI on the new head (7379bb1) is green across the board, including functional tests on lowest deps for PHP 8.1 and 8.3. #352's `InterfaceAliasWiringTest` needed nothing: the only wiring this PR adds is a parameter argument to the already aliased `GiftCardEmailManager`.

The two inline points are separate commits on top: d16a7e2 (an admin send discloses the code) and 7379bb1 ("will be shipped to you"). The description no longer says the PR is stacked on #341, and it has an `## Added in review` section.

PR #345, opened 2026-09-14
Make the 1.x build green again after the merged fixes

loevgaard:
## Summary

The 1.x build turned red after the last merges, and every open PR inherits those failures because PR builds run on the merge with 1.x. Three unrelated causes, all fixed here:

- **Dependency analysis**: `ext-filter` is no longer used (#318 dropped `filter_var()`), so it is removed from `composer.json`; `twig/intl-extra` (added by #331 for `format_date`) is only used from Twig templates, which the analyser does not scan, so it is marked as a known template-only dependency in `composer-dependency-analyser.php`, like `sylius/grid-bundle` and `symfony/translation` already are.
- **Static analysis (lowest deps)**: on the lowest supported Twig, PHPStan rejects the `[RuntimeClass::class, 'method']` callable of a `TwigFilter` exactly as it already did for `TwigFunction` (#321 added the first filter); the existing ignore now covers both.
- **End to end**: the Playwright admin setup intermittently failed with "the stored session does not authenticate" on 1.x (run 34822622428) and on several PRs, passing on rerun. The first requests to a freshly built application race while the dev container and templates compile. The job now warms the cache before serving and waits until the application answers before the first Playwright request.

## Verification

- `composer validate --strict`, `composer normalize --dry-run`, `composer analyse`, unit tests and a YAML lint of the workflow pass locally.
- The dependency analyser only runs in CI (the tool is installed there), so its result is on this PR's checks.


loevgaard (commented):
Reviewed after merge. No issues.

PR #347, opened 2026-09-14
Capture the session files and the logs when the end to end job fails

loevgaard:
## Summary

The end to end job fails now and then with every admin spec after the first reporting `No link matching /\/admin\/gift-cards\/(\d+)$/ found in the grid` (seen on #326, #330, #338 and #320; rerunning the job makes it pass). The Playwright trace of the #320 failure shows the admin session cookie unchanged between the first admin test, which passes, and the second one two seconds later, which is answered with a `302` to the login form and no new session cookie: the server side session lost its security token, or the session file vanished. The suite passes locally every time, and nothing from the server side is captured on CI, so the cause cannot be pinned down from what we have.

This PR captures it, in the workflow only:

- A step before the server starts writes a `session.save_path` pointing into `tests/Application/var/sessions` into an extra ini directory and exports `PHP_INI_SCAN_DIR` (with the leading separator that keeps PHP's own scan directory), so the session files land somewhere the job can upload. The test application leaves session handling to PHP (`handler_id: ~`), so without this the files go wherever the runner's php.ini says.
- The serve step prints PHP's session settings (`save_handler`, `save_path`, `gc_*`, `use_strict_mode`, `lazy_write`), and a step after the suite lists the session files, so the next run shows whether the ini override reached php-fpm.
- On failure the artifact holds the application log (`var/log/dev.log`, debug level, where Symfony's security listeners say why a token was dropped), the session files and the Symfony CLI server log alongside the Playwright report.

No application or plugin code changes. An earlier attempt configured `session.handler.native_file` with an explicit `save_path` in the test application instead; that handler calls `ini_set()` when it is built, which fails in the console once anything has been printed (PHP 8.4's deprecation notices from vendor code do exactly that), so the change stays in CI.

## Verification

- `bin/console lint:yaml .github/workflows/build.yaml` — OK
- The job itself is the verification: its "Serve the test application" and "List the session files" steps show whether the sessions land in the workspace, and the next flaky run ships the logs


.github/workflows/build.yaml
```diff
@@ -278,11 +278,22 @@ jobs:
...
+                    mkdir -p "$GITHUB_WORKSPACE/tests/Application/var/sessions" "$RUNNER_TEMP/php-ini"
+                    printf 'session.save_path=%s\n' "$GITHUB_WORKSPACE/tests/Application/var/sessions" > "$RUNNER_TEMP/php-ini/sessions.ini"
+                    echo "PHP_INI_SCAN_DIR=:$RUNNER_TEMP/php-ini" >> "$GITHUB_ENV"
+
             -   name: "Serve the test application"
                 run: |
                     cd tests/Application
                     symfony serve --no-tls --daemon --port=8080
                     symfony server:status
+                    symfony php -i | grep -E '^session\.(save_handler|save_path|gc_maxlifetime|gc_probability|gc_divisor|use_strict_mode|lazy_write)'
```
loevgaard: Nit: this prints the settings of the command-line PHP, which reads the same scan directory but is not php-fpm. What proves the override reaches php-fpm is the "List the session files" step, and on this run it does.

loevgaard (commented):
Reviewed at 449cd81. Good to merge. The run shows the session files landing in `tests/Application/var/sessions`, and all 48 specs pass. One nit inline.

.github/workflows/build.yaml
loevgaard: Right, `symfony php -i` is the command-line PHP: it reads the extra scan directory, but not php-fpm's php.ini, so the line showed less than the PR description claimed. Fixed in edb2abf by taking the settings from php-fpm itself rather than dropping them: a step of its own, "Show the session settings php-fpm reads", runs `php-fpm8.3 -i` (the binary `server:status` names) before the server starts. On the run after the push it reports `Loaded Configuration File => /etc/php/8.3/fpm/php.ini`, not the command line's, with `save_path` in the workspace and `gc_probability => 0`.

It is still that binary run on its own, not the running pool, which the Symfony CLI starts with its own pool configuration on top. The step's comment says so and names "List the session files" as what proves the override reaches the server; on that run it lists 18 session files. I kept a settings step rather than dropping it because php-fpm's own GC settings are what I'd want to read next to a session file that vanished.

PR #349, opened 2026-09-15
Warn in the admin when a channel sells gift cards without an enabled design

### LINKED ISSUE #348: Warn in the admin when a channel sells gift cards but has no enabled design
Follow-up to #325 (which fixes #285). The default design is no longer created on the fly while a product page renders: `GiftCardDesignProvider` only reads, the `classic` design comes from the `setono_gift_card_design` fixture or from `setono:gift-card:create-default-design`, and a channel without an enabled design simply gets no design picker on the product page (the card then renders its framed default).

That is a working state, but a merchant who forgot the setup step will not notice: the product page just looks plainer than intended, and physical cards ship without artwork. The design picker and the live preview are what make the product page feel like buying a card, so the missing setup should be visible in the admin rather than found by customers.

### Proposal

- A checker returning the channels that have an enabled gift card product but no enabled design (`GiftCardDesignRepositoryInterface::findEnabledByChannel()` and the product's `isGiftCard()` flag are there). Only that combination hurts; a channel that does not sell gift cards, or a shop that sells virtual cards only and deliberately has no designs, must not be nagged.
- A short warning in the admin top bar on every page, rendered through the `sylius.admin.layout.topbar_middle` template event (the only layout event Sylius 1.x exposes across all admin pages; the content area has none and overriding `@SyliusAdmin/layout.html.twig` would collide with hosts' own overrides). It names the channels and links to Gift cards → Designs.
- A full message box on the gift card designs index and the gift cards index, which the plugin owns, explaining the two ways to fix it: create a design there, or run `setono:gift-card:create-default-design`.
- One cheap query per admin page; no configuration flag as long as the check is scoped as above.

### Acceptance criteria

- With the fixtures loaded, no warning is shown anywhere in the admin.
- Disable every design of a channel that has a gift card product: the top bar warning names that channel on every admin page and the designs index shows the message box with both remedies.
- Remove the gift card product from that channel (or disable it): the warning disappears without any design being created.
- Playwright coverage for the warning appearing and disappearing, unit coverage for the checker.

loevgaard:
Stacked on #325: the default design is no longer created on the fly, so a channel can sell gift cards with nothing to print them with. The product page copes (no design picker, the framed default card), but the merchant would only find out from customers. This makes the state visible in the admin.

Fixes #348

## Changes

- `Checker\GiftCardSetupChecker` (+ interface): the enabled channels that have an enabled gift card product but no enabled gift card design. The product flag lives on the host's product class, so the channels selling gift cards come from one DQL query on `%sylius.model.product.class%`; the designs from `GiftCardDesignRepositoryInterface::findEnabledByChannel()`. A channel that does not sell gift cards is never named, so a shop selling virtual cards only and deliberately running without designs is not nagged.
- `Twig\Runtime\GiftCardSetupRuntime` / `GiftCardSetupExtension`: `setono_gift_card_channels_without_design()`, memoised per request because the top bar and the index message both ask on the same page. One query per admin page.
- `sylius_ui` events prepended by the extension: a label in `sylius.admin.layout.topbar_middle` (the only layout event Sylius 1.x exposes on every admin page; the content area has none, and overriding `@SyliusAdmin/layout.html.twig` would collide with hosts' own overrides) naming the channels and linking to the designs index; a `ui warning message` above the header of `setono_sylius_gift_card.admin.gift_card.index` and `…gift_card_design.index` (priority 30, ahead of Sylius' header at 20) with the two ways out: a link to create a design, or `setono:gift-card:create-default-design`.
- Translations in en, da, fr under `setono_sylius_gift_card.ui.setup_warning.*`; a README paragraph next to the command.
- Tests: `tests/Functional/GiftCardSetupCheckerTest.php` walks through every combination (no gift card product, disabled product, ordinary product, disabled design, design in another channel, two channels); `tests/Unit/Twig/Runtime/GiftCardSetupRuntimeTest.php` pins the memoisation; `tests/Playwright/specs/admin/setup-warning.spec.js` asserts nothing is shown on the seeded shop, then disables every design, checks the label on the products page, follows it to the designs index and its message, checks the gift cards index too, and re-enables the designs in a `finally` so later specs find the shop as seeded.

## Verification

From the worktree on PHP 8.4, own database and port.

- `vendor/bin/phpunit --testsuite unit` — OK (137 tests, 466 assertions)
- `vendor/bin/phpunit --testsuite functional` — OK (47 tests, 216 assertions)
- `composer analyse`, `composer check-style`, `vendor/bin/rector process --dry-run` — all `[OK]`
- `lint:container`, `lint:twig` (22 files), `lint:yaml` (14 files) — all `[OK]`
- `npx playwright test` on fresh fixtures — 47 passed, including the new spec

## Notes for reviewers

- The check runs on every admin page. It is two cheap queries (a `DISTINCT` on product channels and one per channel that sells gift cards), which seemed fine without a cache or a configuration flag; if it ever shows up in a profile, caching the answer for a few minutes in the runtime is a small change.
- The label sits in the top bar's middle slot, which Sylius leaves empty, as an `ui orange label` inside a menu item so it reads as a warning without a stylesheet of its own.


src/Twig/Runtime/GiftCardSetupRuntime.php
```diff
@@ -0,0 +1,32 @@
...
+    public function __construct(private readonly GiftCardSetupCheckerInterface $setupChecker)
+    {
+    }
+
+    /**
+     * @return list<ChannelInterface>
+     */
+    public function getChannelsWithoutDesign(): array
+    {
+        return $this->channelsWithoutDesign ??= $this->setupChecker->getChannelsWithoutDesign();
```
loevgaard: The runtime is a shared service, so under a worker runtime (FrankenPHP worker mode, RoadRunner) this cache outlives the request: once computed, the warning neither appears nor goes away until the worker restarts. Implementing `ResetInterface` and tagging the service with `kernel.reset` keeps it per request. The XML config does not autoconfigure, so the tag has to be explicit.

loevgaard (commented):
Reviewed at cddf389. Good to merge apart from the nit inline. I checked that `sylius.admin.layout.topbar_middle` exists in Sylius' admin layout, and the checker only names enabled channels that sell an enabled gift card product.

src/Twig/Runtime/GiftCardSetupRuntime.php
loevgaard: Agreed, done in cdea833. `GiftCardSetupRuntime` implements `ResetInterface`, and `reset()` clears the memoised channels. The service in `services/twig.xml` carries `<tag name="kernel.reset" method="reset"/>` explicitly, with a comment saying why, and `debug:container --tag=kernel.reset` in the test application lists it. The plugin now requires `symfony/service-contracts` (`^2.5 || ^3.0`, the constraint Symfony's own components use) directly, since it uses one of its interfaces.

`GiftCardSetupRuntimeTest::it_forgets_the_answer_when_reset` pins that the checker is asked again after `reset()` and that the new answer is then memoised. Because nothing is autoconfigured, `SetonoSyliusGiftCardExtensionTest::it_tags_every_resettable_service_for_the_kernel_to_reset` also loads the extension and fails for any plugin service that implements `ResetInterface` without the tag. Without the tag it fails, naming the runtime. The runtime still takes the checker through `GiftCardSetupCheckerInterface`, and `InterfaceAliasWiringTest` passes on the rebased branch.

